Issue #43, opened 2020-07-14
Install required version of Pyshacl when installing Nexus Forge


annakristinkaufmann:
The `Shape' object has no attribute 'traverse'` error described here: #36 is due to the installed version of `Pyshacl`. On 2020-07-10, a new version of `Pyshacl` was released on PyPi and `Nexus Forge` does not specify a particular `Pyshacl` version for installation. I.e., when installing the Nexus Forge now, the latest release `Pyshacl 0.12.0` will be installed which causes the error. Running the same notebook as in #36  with `Pyshacl 0.11.6.post1` works fine.

MFSY:
@annakristinkaufmann is this fixed for now ?

annakristinkaufmann:
Yes!

PR #44, opened 2020-07-15
Fix temporary the Shape.traverse() issue

pafonta:
See #43.

The actual fix is about what happens at: https://github.com/BlueBrain/nexus-forge/blob/1f1c47b661769f1f9ce871b804db98d680307de9/kgforge/specializations/models/rdf/service.py#L110.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -47,7 +47,7 @@ setup(
         "aiohttp",
         "nest_asyncio",
         "pyLD",
-        "pyshacl",
+        "pyshacl==0.11.6.post1",
         "rdflib-jsonld"
     ],
     extras_require={


setup.py
```diff
@@ -47,7 +47,7 @@
         "aiohttp",
         "nest_asyncio",
         "pyLD",
-        "pyshacl",
+        "pyshacl==0.11.6.post1",
```
MFSY, line 50: Can we use [v0.11.6](https://github.com/RDFLib/pySHACL/releases/tag/v0.11.6) instead of [v0.11.6.post1](https://github.com/RDFLib/pySHACL/releases/tag/v0.11.6.post1) ?

@annakristinkaufmann Did you try with v0.11.6 too as part of #44

annakristinkaufmann: @MFSY Yes, v0.11.6 worked as well as far as I remember.

pafonta: Why using an older version?

MFSY: v0.11.6 looks newer (even though both releasing were made the same day) from the tag listing: https://github.com/RDFLib/pySHACL/releases

annakristinkaufmann: I think v0.11.6.post1 is newer: https://github.com/RDFLib/pySHACL/compare/v0.11.6...v0.11.6.post1

MFSY: Okay. The tag listing is misleading then.

PR #45, opened 2020-07-16
Prevent Store.download() to overwrite existing files

pafonta:
Improvements when using `forge.download(...)` and `Dataset.download(...)`:
- If the given directory path doesn't exist and no URLs were found, the path is not created.
- When a file with the same file name is already existing, it is not overwritten.
The concerned downloaded file has a timestamp added as suffix.
Here is an example:
store-download-no_overwriting

### DIFF
--- a/kgforge/core/archetypes/store.py
+++ b/kgforge/core/archetypes/store.py
@@ -132,7 +132,8 @@ class Store(ABC):
         # TODO This operation might be adapted if other file metadata are needed.
         not_supported()
 
-    def download(self, data: Union[Resource, List[Resource]], follow: str, path: str) -> None:
+    def download(self, data: Union[Resource, List[Resource]], follow: str, path: str,
+                 overwrite: bool) -> None:
         # path: DirPath.
         urls = collect_values(data, follow, DownloadingError)
         count = len(urls)
@@ -145,7 +146,10 @@ class Store(ABC):
         for x in urls:
             filename = self._retrieve_filename(x)
             filepath = dirpath / filename
-            filepaths.append(f"{filepath}.{timestamp}" if filepath.exists() else str(filepath))
+            if not overwrite and filepath.exists():
+                filepaths.append(f"{filepath}.{timestamp}")
+            else:
+                filepaths.append(str(filepath))
         if count > 1:
             self._download_many(urls, filepaths)
         else:
--- a/kgforge/core/archetypes/store.py
+++ b/kgforge/core/archetypes/store.py
@@ -135,6 +135,9 @@ class Store(ABC):
     def download(self, data: Union[Resource, List[Resource]], follow: str, path: str) -> None:
         # path: DirPath.
         urls = collect_values(data, follow, DownloadingError)
+        count = len(urls)
+        if count == 0:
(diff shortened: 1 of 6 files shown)


pafonta:
Hello @MFSY,

Would you have had time to look at this PR?

Thank you.

kgforge/specializations/stores/bluebrain_nexus.py
```diff
@@ -169,7 +169,16 @@ def retrieve(self, id: str, version: Optional[Union[int, str]] = None) -> Resour
             self.service.sync_metadata(resource, response)
             return resource
 
-    def _download_one(self, url: str, path: Path) -> None:
+    def _retrieve_filename(self, id: str) -> str:
```
MFSY, line 172: The filename corresponds to the distribution.name value. Can't that be used instead of getting the file metadata from Nexus ?

kgforge/core/archetypes/store.py
```diff
@@ -127,28 +128,38 @@ def retrieve(self, id: str, version: Optional[Union[int, str]]) -> Resource:
         # TODO These two operations might be abstracted here when other stores will be implemented.
         pass
 
+    def _retrieve_filename(self, id: str) -> str:
+        # TODO This operation might be adapted if other file metadata are needed.
+        not_supported()
+
     def download(self, data: Union[Resource, List[Resource]], follow: str, path: str) -> None:
```
MFSY, line 135: It would be nice to introduce a choice for a user to still erase if his/her intent is to only have the latest download.

kgforge/specializations/stores/bluebrain_nexus.py
pafonta: There is no guarantee that this is the case. Indeed, someone could decide to model their data differently.

This implementation offers:
1. independence with how people model their data,
2. consistency across different Store implementations.

kgforge/core/archetypes/store.py
pafonta: Should it be done in this PR?

pafonta:
> I did not see a test coming with this PR. It would be great to have a test for this case.

Indeed. However, this needs either the implementation of file operations for DemoStore or a mock of Nexus. The two tasks are still in the backlog. This therefore cannot be done at the moment.

kgforge/specializations/stores/bluebrain_nexus.py
MFSY: Okay. Let proceed with your version.

kgforge/core/archetypes/store.py
MFSY: In this one or in another one.

MFSY:
> Indeed. However, this needs either the implementation of file operations for DemoStore or a mock of Nexus. The two tasks are still in the backlog. This therefore cannot be done at the moment.

Okay. Let proceed without them for now.

kgforge/core/archetypes/store.py
pafonta: Added in this PR as commit 4a54135.

pafonta:
Hello @MFSY,

As discussed, I am merging this PR.

PR #39, opened 2020-07-10
travis.yml to release on PyPI

pafonta:

PR #46, opened 2020-07-20
Adapted template test data

MFSY:
Adapted template test data to align with changes introduced in 95f69860bb39df8a749605da56665c9d108089b8
- model._template(type_, False) output 'id' when nodeeKind:IRI is enforced in the SHACL shape
- Templates are nested when a property shape has a 'sh:node' constraint or a type for which a shape was defined

Change-Id: I78df27b3c9c85ab234c5ecd9b3842ea1899348d9

### DIFF
--- a/kgforge/specializations/models/rdf/service.py
+++ b/kgforge/specializations/models/rdf/service.py
@@ -196,7 +196,7 @@ class RdfService:
             l_terms = dict()
             for property_ in properties:
                 if hasattr(property_, "path"):
-                    if property_.path != RDF.type:
+                    if property_.path != RDF.type and str(property_.path) != "id":
                         v_prefix, v_namespace, v_name = self._graph.compute_qname(property_.path)
                         l_prefixes.update({v_prefix: str(v_namespace)})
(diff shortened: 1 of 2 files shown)


pafonta:
Hello @MFSY,

I am not sure why https://github.com/BlueBrain/nexus-forge/commit/95f69860bb39df8a749605da56665c9d108089b8 introduced the need for
```
"postalCode": "",
```
to now be
```
"postalCode": ["",0],
```
but the tests are passing so this PR could be merge to unblock https://github.com/BlueBrain/nexus-forge/pull/39.

MFSY:
The reason is because of the SHACL 'OR' clause. The [postal address shape](https://github.com/BlueBrain/nexus-forge/blob/1f1c47b661769f1f9ce871b804db98d680307de9/tests/data/shacl-model/commons/shapes-1.json#L89) is expecting a string or an integer. One way of generating the template is to bring all possible default values (here "" for string and 0 for integer). 

We'll definitely need to have a fresh look at this.

PR #47, opened 2020-07-23
Implement bulk upload and bulk download

pafonta:
Known issue:
Filenames are prefixed with `utf-8''`.
It might be because `filename*` (with a star), used by `aiohttp`, is not handled by Nexus.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,29 @@
+language: python
+
+python:
+  - "3.6"
+  - "3.7"
+  - "3.8"
+
+install:
+  - pip install .[dev]
+
+script:
+  - pytest --cov=./
+
+after_success:
+  - codecov
+
+branches:
+  only:
+    - master
+
+deploy:
+  provider: pypi
+  on:
+    tags: true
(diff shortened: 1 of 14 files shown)


pafonta:
The issue with prefixed filenames is fixed in https://github.com/BlueBrain/nexus-forge/pull/47/commits/28aca88980eb1e7476c4f44062074f5ab19f9dda.

The issue seems caused by Nexus not parsing the `Content-Disposition` `filename*` properly.

The hack is to prevent sending `Content-Disposition` with `filename*`.

I commented the non hacky way using `aiohttp.FormData` and added a `FIXME` in the code.

With the hack, **non ASCII filenames are not handled** by the Forge when using the store Nexus.

From my understanding, `aiohttp` sends with `FormData` a `Content-Disposition` with `filename` and `filename*` but Nexus is using the encoded value of `filename*` as a value for `filename` which leads to ending with filenames prefixed with their encoding.

pafonta:
Hello @MFSY,

As discussed, I am merging this PR.

PR #48, opened 2020-07-23
Use codecov.io for code coverage reports

pafonta:


MFSY:
This Travis CI Branch check looks confusing. From what I understand [here](https://travis-ci.community/t/travis-ci-branch-expected-waiting-for-status-to-be-reported/829), it is building and checking the downstream branch. If that branch does not have travis configured then it waits for status forever.

Should we keep it ?

pafonta:
Hello @MFSY,

This isn't related to this PR or codecov.io.

> Should we keep it ?

I would say no by deactivating the need for this check to pass as said [there](https://travis-ci.community/t/travis-ci-branch-expected-waiting-for-status-to-be-reported/829/2).

MFSY:
Okay. I removed it for now. But we'll need to investigate to know more about this.

PR #121, opened 2021-03-15
Added atLocation property to file-to-resource-mapping.hjson for nexus store

annakristinkaufmann:
To enable direct access to the storage location of a file through the distribution object, the property **atLocation** was added to the file-to-resource.hjson mapping for the Nexus store.

This property provides the @id of the store in which a file is stored. Additionally, it provides the **location** of the file in cases where a remote disk storage is being used as store.


MFSY (approved):

MFSY:
Thanks @annakristinkaufmann . The file location can be retrieve from a dataset like in the following:

`dataset.distribution.atLocation.location`

PR #210, opened 2021-12-15
Add support for returning resolvers as dictionary

kplatis:


kgforge/core/forge.py
```diff
@@ -224,13 +224,47 @@ def validate(self, data: Union[Resource, List[Resource]],
...
     # Resolving User Interface.
 
-    def resolvers(self):
-        print("Available scopes:")
-        for k, v in sorted(self._resolvers.items()):
-            print(" - ", k, ":")
-            for r_key, r_value in v.items():
-                print("     - resolver: ", r_key)
-                print("         - targets: ",  ",".join(r_value.targets.keys()))
+    def resolvers(self, as_dict: bool = False) -> Optional[Dict]:
```
MFSY, line 227: forge.template() uses `output: str` to control the type of output. I recommend to reuse 'output' instead of 'as_dict'

kgforge/core/forge.py
```diff
@@ -224,13 +224,47 @@ def validate(self, data: Union[Resource, List[Resource]],
...
+        """
+        Prints or returns the available resolvers of the graph.
```
MFSY, line 229: I would remove '...of the graph part'.

kgforge/core/forge.py
```diff
@@ -224,13 +224,47 @@ def validate(self, data: Union[Resource, List[Resource]],
...
+
+        If as dict is False (default), prints the available scopes
+        If as_dict is True, returns a dictionary of the available scores and their targets
```
MFSY, line 232: Minor typo: 'scopes' here

tests/core/test_resolving.py
```diff
@@ -13,3 +13,22 @@
 # along with Blue Brain Nexus Forge. If not, see <https://choosealicense.com/licenses/lgpl-3.0/>.
 
 # Placeholder for the test suite for selection of target or resolver from input and configuration.
+
+from kgforge.core import KnowledgeGraphForge
```
MFSY, line 17: Should'nt this go to tests/core/test_forge.py where forge init functions are done. test_resolving is more for the actual resolving business.

tests/core/test_resolving.py
```diff
@@ -13,3 +13,22 @@
...
+    Tests the .resolver(as_dict=True) function that returns a dictionary
+    """
+
+    def test_resolver_returns_dict(self, config):
+        forge = KnowledgeGraphForge(config)
+        resolvers_dict = forge.resolvers(as_dict=True)
+        assert type(resolvers_dict) is dict
+
+    def test_resolver_returns_correct_object(self, config):
+        dict_result = {'terms': {'sex': {'source': 'sex.json'}}}
```
MFSY, line 31: Looking again at the forge config, I think we should use 'bucket' instead of 'source' to avoid clashes with the actual source defined in the forge config.

MFSY (commented):
Nice @kplatis ,

I have very few comments.

kgforge/core/forge.py
kplatis: So in that case `output` should be by default `True` in order to keep backwards compatibility with current implementation and when it is False the function should return the dictionary instead of printing.

kplatis: I just noticed that in your comment you mention `output` to be `string`. How do you think of it being used?

MFSY: It is used as in: https://github.com/BlueBrain/nexus-forge/blob/b60fe2163db5fa331e7edb3bf2e5f86381c5cf0a/kgforge/core/archetypes/model.py#L93

kplatis: I modified the implementation to accept string instead of bool and pushed the changes

kgforge/core/forge.py
kplatis: I removed it and pushed the changes

kgforge/core/forge.py
kplatis: I changed the pydoc and pushed the changes

tests/core/test_resolving.py
kplatis: I moved the unit test to test_forge.py and pushed the changes

tests/core/test_resolving.py
kplatis: I changed the key and pushed the changes

MFSY (approved):

PR #217, opened 2022-01-10
Add support for validation against a schema

### LINKED ISSUE #207: Add support for validation against a schema
The current implementation of `.validate()` expects a `type` attribute in the validated resource object. It would be useful to be able to validate against a schema by passing the schema as parameter in `.validate()`

kplatis:
Fixes #207

Added a `schema` parameter in `.validate()` function. The resource is validated against the provided schema


kgforge/specializations/models/rdf/service.py
```diff
@@ -155,9 +155,12 @@ def materialize(self, iri: URIRef) -> NodeProperties:
...
         raise NotImplementedError()
 
-    def validate(self, resource: Resource):
+    def validate(self, resource: Resource, schema: str=None):
         try:
-            shape_iri = self.types_to_shapes[resource.type]
+            if schema is None:
+                shape_iri = self.types_to_shapes[resource.type]
+            else:
+                shape_iri = self.types_to_shapes[schema]
```
MFSY, line 163: Looking at this, it looks this implementation is expecting the value of 'schema' to be a type rather than an actual SHACL schema id. I recommend to rename it to `type` to reflect that and differentiate with `schema_id` argument in  [forge.py#L630](https://github.com/BlueBrain/nexus-forge/blob/0b04e7a1c8d6d632c8d4b2d8db6e65b79387df3d/kgforge/core/forge.py#L630) and [forge.py#L642](https://github.com/BlueBrain/nexus-forge/blob/0b04e7a1c8d6d632c8d4b2d8db6e65b79387df3d/kgforge/core/forge.py#L642).

Also a resource can have multiple types (i.e. resource.type=["type1", "type2"]), in such case the `type` argument can help the user specify which one to use. In case of multiple types and without the `type` argument provided, then a value error should be raised for now. We can extend later to support validating using multiple types.

kgforge/specializations/models/demo_model.py
```diff
@@ -81,7 +81,7 @@ def mapping(self, entity: str, source: str, type: Callable) -> Mapping:
 
     # Validation.
 
-    def _validate_one(self, resource: Resource) -> None:
+    def _validate_one(self, resource: Resource, **kwargs) -> None:
```
MFSY, line 84: Why `**kwargs` here and not a type argument here as in rdf_model ?

I think we should have a parallel between demo_model and rdf_model w.r.t this `_validate_one` method.

kgforge/specializations/models/rdf/service.py
kplatis: @MFSY  I get it. However the keyword `type` shadows the `type` built-in of Python. Is `type_id` a good alternative?

MFSY: True. Let go for `type_`. We will need later to propagate this to some forge public methods (which uses `type`) while keeping backward compatibility

kplatis: @MFSY 
> True. Let go for type_.

I pushed the change for that

> Also a resource can have multiple types (i.e. resource.type=["type1", "type2"]), in such case the type argument can help the user specify which one to use.

I don't quite understand this part. Doesn't `type_` override the `resource.type` attribute in `validate`?

kgforge/specializations/models/demo_model.py
kplatis: I resolved it in the last push

kgforge/specializations/models/rdf/service.py
MFSY: >I don't quite understand this part. Doesn't type_ override the resource.type attribute in validate?

Yes.

kgforge/specializations/models/demo_model.py
```diff
@@ -81,7 +81,7 @@ def mapping(self, entity: str, source: str, type: Callable) -> Mapping:
 
     # Validation.
 
-    def _validate_one(self, resource: Resource) -> None:
+    def _validate_one(self, resource: Resource, type_: str=None) -> None:
         # If the resource is not typed, AttributeError is raised: run() sets _validated to False.
         type_ = resource.type
```
MFSY, line 86: the 'type_' argument should be used.

kgforge/specializations/models/demo_model.py
```diff
@@ -81,7 +81,7 @@ def mapping(self, entity: str, source: str, type: Callable) -> Mapping:
 
     # Validation.
 
-    def _validate_one(self, resource: Resource) -> None:
+    def _validate_one(self, resource: Resource, type_: str=None) -> None:
         # If the resource is not typed, AttributeError is raised: run() sets _validated to False.
```
MFSY, line 85: This text should be adapted to mention 'type_'

kgforge/core/forge.py
```diff
@@ -283,7 +283,8 @@ def template(
     def validate(
         self,
         data: Union[Resource, List[Resource]],
-        execute_actions_before: bool = False,
+        execute_actions_before: bool=False,
+        type_=None
```
MFSY, line 287: I recommend to have 'None' as a default value only here at forge level. The downstream methods can just have 'type_' without default value.

kgforge/specializations/models/rdf/service.py
MFSY: My point was more about [this line](https://github.com/BlueBrain/nexus-forge/pull/217/files#diff-20fe56168eb41def80c78aa3d4c1f0dbeddb920f8859a795c86f3f148cc4c408R161) which will throw an error when resource.type is a list. A strategy (in case no type_ is provided) could be to check whether resource.type is a list and throw an exception (ValueError may be ?) with a meaningful error message.

kgforge/core/forge.py
kplatis: @MFSY By downstream methods you mean the methods that are not in `forge.py`?

MFSY: Yes.  this method for example call a private method from _model.

kgforge/specializations/models/rdf/service.py
kplatis: @MFSY I pushed a change that checks in order to raise ValueError

kgforge/specializations/models/demo_model.py
kplatis: I pushed a change for that

kgforge/specializations/models/demo_model.py
kplatis: I pushed a change that changes the text

kgforge/core/forge.py
kplatis: I changed the defaults to be None only for `.validate()` in `forge.py`. All other methods get the value downstream from `validate`

PR #265, opened 2022-08-16
Support content_type when downloading

### LINKED ISSUE #251: Support content_type when calling forge.download and Dataset.download

MFSY:
* Support downloading data of a given content-type
*  When downloading, set the default json property path to follow to "distribution.contentUrl" in order to get the 
downloadable file urls and the default download path to the current folder
* Disable rewriting following SPARQL keywords: "as", "not", "exists", "group", "by"
* Updated SPARQL query statement builder to consider @id and @type values as URIs when used with the NOT_EQUAL operator
* Added methods to  check if a Resource has an identifier or a type and added methods to get them

Fixes #251


crisely09:
Oh seems like I rebased in two places, that is why the added (unnecessary) commits, I am sorry about tat mess.

kgforge/specializations/stores/bluebrain_nexus.py
```diff
@@ -209,18 +210,23 @@ def _register_one(self, resource: Resource, schema_id: str) -> None:
...
             params_register = copy.deepcopy(self.service.params.get("register", None))
-            response = requests.post(
-                url,
-                headers=self.service.headers,
-                data=json.dumps(data, ensure_ascii=True),
-                params=params_register,
-            )
+            if resource.has_identifier()[0]:
+                url = f"{url_base}/{quote_plus(resource.get_identifier())}"
+                response = requests.put(
```
crisely09, line 216: Maybe you can explain why you use `get.put` instead of `get.post`.

kgforge/specializations/stores/bluebrain_nexus.py
```diff
@@ -209,18 +210,23 @@ def _register_one(self, resource: Resource, schema_id: str) -> None:
...
-                else url_base
-            )
             params_register = copy.deepcopy(self.service.params.get("register", None))
-            response = requests.post(
-                url,
-                headers=self.service.headers,
-                data=json.dumps(data, ensure_ascii=True),
-                params=params_register,
-            )
+            if resource.has_identifier()[0]:
```
crisely09, line 214: I would do: 

```suggestion
            identifier = resource.get_identifier()
            if identifier:
                url = f"{url_base}/{quote_plus(identifier)}"
```

kgforge/core/resource.py
```diff
@@ -95,6 +95,36 @@ def _set_synchronized(self, sync: bool) -> None:
     
     _synchronized = property(_get_synchronized, _set_synchronized)
 
+    def has_identifier(self, return_attribute: bool = False) ->  Union[bool, Tuple[bool, str]]:
```
crisely09, line 98:
```suggestion
    def has_term(self, term, return_attribute: bool = False) ->  Union[bool, Tuple[bool, str]]:
            result = False
            attribute = None
            at_term = f"@{term}"
            if hasattr(self, at_term):
                result : True
                attribute = at_term
            if hasattr(self, term):
                result: True
                attribute = term
            return result, attribute if return_attribute else result
```

crisely09: To replace has_type or has_identifier, and have a single method for any term. Of course, one would need to update the other methods: the `get_type` method will pass `term=type` and `get_identifier` `term=id`.

kgforge/specializations/stores/bluebrain_nexus.py
```diff
@@ -391,11 +397,13 @@ def _download_many(
         paths: List[str],
         store_metadata: Optional[DictWrapper],
         cross_bucket: bool,
+        content_type: str
     ) -> None:
         async def _bulk():
             loop = asyncio.get_event_loop()
             semaphore = Semaphore(self.service.max_connection)
-            async with ClientSession(headers=self.service.headers_download) as session:
+            headers = update_dict(self.service.headers_download, {"Accept":content_type})
```
crisely09, line 405: This should only be executed when `content_type` is NOT `None`. Because the `forge.download` method passes `None` as default. So it gives the error `request is missing required HTTP header 'Accept'` when `content_type` is not given

kgforge/specializations/stores/bluebrain_nexus.py
MFSY: PUT verb is used here because there is an id provide when creating the resource. POST is more for collection (i.e you are not targeting specific resource)

kgforge/specializations/stores/bluebrain_nexus.py
```diff
@@ -209,18 +210,23 @@ def _register_one(self, resource: Resource, schema_id: str) -> None:
...
-            )
             params_register = copy.deepcopy(self.service.params.get("register", None))
-            response = requests.post(
-                url,
-                headers=self.service.headers,
-                data=json.dumps(data, ensure_ascii=True),
-                params=params_register,
-            )
+            if resource.has_identifier()[0]:
+                url = f"{url_base}/{quote_plus(resource.get_identifier())}"
```
crisely09, line 217: This is now repeated, the previous line does the same, according to the change.

MFSY: This is fixed now

kgforge/specializations/stores/bluebrain_nexus.py
MFSY: This is fixed

crisely09:
Looks good to me.

crisely09 (approved):

Issue #270, opened 2022-09-22
Retrieve a resource by _self


MFSY:
Currently forge.retrieve(id=resource.id) only accepts resource.id when using BlueBrainNexus. Under the hood forge is building a Nexus address for the requested URI by concatenating the endpoint, the org, the proj, the schema and then the URI. A resource._store_metadata._self is already made of the endpoint, the org, the proj, the schema and then the URI.

forge.retrieve(id=resource._store_metadata._self) should work

ssssarah:
Can we close this?

PR #271, opened 2022-09-26
Small change to use _self value as id when retrieving a resource.

crisely09:
This is a veeeery simple approach that achieves the desired behavior in https://github.com/BlueBrain/nexus-forge/issues/270

Unit tests are missing at the moment.


kgforge/specializations/stores/bluebrain_nexus.py
```diff
@@ -351,21 +351,38 @@ def retrieve(
...
             )
             response.raise_for_status()
+        except HTTPError as er:
+            nexus_path = f"{self.service.endpoint}/resources/"
+            if not cross_bucket:
+                nexus_path += f"{self.service.organisation}/{self.service.project}"
+            # Try to use the id as it was given
+            if url_base in id_without_query or nexus_path in id:
+                # Use the given id in the request
+                url = id_without_query
```
MFSY, line 361: I would check the retrieve_source so that to add /source:

`if retrieve_source and not cross_bucket:
                    url = "/".join((id_without_query, "source"))
                else:
                    url = id_without_query`

kgforge/specializations/stores/bluebrain_nexus.py
```diff
@@ -351,21 +351,38 @@ def retrieve(
...
+                raise RetrievalError(_error_message(er))
+        finally:
             if retrieve_source and not cross_bucket:
 
                 response_metadata = requests.get(
                     url_resource, params=query_params, headers=self.service.headers
                 )
                 response_metadata.raise_for_status()
-            if retrieve_source and cross_bucket:
+            elif retrieve_source and cross_bucket:
```
MFSY, line 378: I would add ` and response and "_self" in response.json()` to make response.json() has _self

kgforge/specializations/stores/bluebrain_nexus.py
```diff
@@ -351,21 +351,38 @@ def retrieve(
...
                     "/".join([response.json()["_self"], "source"]), params=query_params, headers=self.service.headers
                 )
                 response_metadata.raise_for_status()
-
-        except HTTPError as e:
-            raise RetrievalError(_error_message(e))
-        else:
+            else:
+                response_metadata = True # when retrieve_source is False
+        if response_metadata:
```
MFSY, line 385: I would add `and response`

kgforge/specializations/stores/bluebrain_nexus.py
```diff
@@ -351,21 +351,38 @@ def retrieve(
                 url, params=query_params, headers=self.service.headers
             )
             response.raise_for_status()
+        except HTTPError as er:
+            nexus_path = f"{self.service.endpoint}/resources/"
+            if not cross_bucket:
+                nexus_path += f"{self.service.organisation}/{self.service.project}"
+            # Try to use the id as it was given
+            if url_base in id_without_query or nexus_path in id:
```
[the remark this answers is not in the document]
crisely09: done

kgforge/specializations/stores/bluebrain_nexus.py
crisely09: In fact, I think the /source does not exist, and the id_without_query should be used directly.

kgforge/specializations/stores/bluebrain_nexus.py
```diff
@@ -351,21 +351,38 @@ def retrieve(
...
+                    response.raise_for_status()
+                except HTTPError as e:
+                    raise RetrievalError(_error_message(e))
+            else:
+                raise RetrievalError(_error_message(er))
+        finally:
             if retrieve_source and not cross_bucket:
 
                 response_metadata = requests.get(
                     url_resource, params=query_params, headers=self.service.headers
```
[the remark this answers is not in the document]
crisely09: I added `url_resource = id_without_query` instead, because if not it could break the case where the id was given.

kgforge/specializations/stores/bluebrain_nexus.py
crisely09: done

kgforge/specializations/stores/bluebrain_nexus.py
crisely09: done

kgforge/specializations/stores/bluebrain_nexus.py
MFSY: What do you mean by `... /source does not exist...` ?
If retrieve_source is true, the requested url should be suffixed with `/source` in order to get the resource payload as the user push it. Actually the same thing as in [L346](https://github.com/BlueBrain/nexus-forge/pull/271/files#diff-9ee1d07daf4890e0d1087d9815baaea67a26d4fd6b31e546784b695165c8e68aR346).

That '/source' missing is the reason why the two resources (retrieved from id and url) are not equal in the section `Retrieving using the resorce url` section of the [`04  - Querying.ipynb`](https://github.com/BlueBrain/nexus-forge/blob/c203996ccf162943e05ba34a0c308e9b9c46180e/examples/notebooks/getting-started/04%20-%20Querying.ipynb) notebook

kgforge/specializations/stores/bluebrain_nexus.py
```diff
@@ -1018,6 +1043,63 @@ def format_message(msg):
         return format_message(str(error))
 
 
+def build_sparql_query_statements(context: Context, *conditions) -> Tuple[List, List]:
```
MFSY, line 1046: This was ported in [SPARQLQueryBuilder](https://github.com/BlueBrain/nexus-forge/blob/master/kgforge/core/commons/sparql_query_builder.py).

kgforge/specializations/stores/bluebrain_nexus.py
```diff
@@ -351,21 +352,41 @@ def retrieve(
...
                 response_metadata = requests.get(
                     "/".join([response.json()["_self"], "source"]), params=query_params, headers=self.service.headers
                 )
                 response_metadata.raise_for_status()
-
-        except HTTPError as e:
-            raise RetrievalError(_error_message(e))
-        else:
+            else:
+                print("retrieve_source", retrieve_source)
```
MFSY, line 385: Remove print

kgforge/specializations/stores/bluebrain_nexus.py
crisely09: Oh, I think you are correct. I did the change and now it works fine.

kgforge/specializations/stores/bluebrain_nexus.py
```diff
@@ -351,21 +352,41 @@ def retrieve(
                 url, params=query_params, headers=self.service.headers
             )
             response.raise_for_status()
+        except HTTPError as er:
+            nexus_path = f"{self.service.endpoint}/resources/"
```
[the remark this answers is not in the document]
crisely09: yes, this was old, I removed it

kgforge/specializations/stores/bluebrain_nexus.py
crisely09: Ah, yes, I think it got in by mistake when I rebased.

kgforge/specializations/stores/bluebrain_nexus.py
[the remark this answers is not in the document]
MFSY: > because if not it could break the case where the id was given.
Works fine when I try. 
If `id_without_query ` is not used here then:
* forge.retrieve(id=_self, cross_bucket=False, retrieve_source=True) will not retrieve any _store_metadata because /source is always used in try, except and finally clauses. Actually we should always have two calls once without /source and one with to get _store_metadata back.

kgforge/specializations/stores/bluebrain_nexus.py
```diff
@@ -351,21 +352,38 @@ def retrieve(
                 url, params=query_params, headers=self.service.headers
             )
             response.raise_for_status()
+        except HTTPError as er:
+            # Try to use the id as it was given
+            if id.startswith(self.service.url_resources):
```
MFSY, line 357: So after few more tests, I think the nexus_path you had is needed for cross_bucket=True:

`nexus_path = f"{self.service.endpoint}/resources/"
          if not cross_bucket:
              nexus_path += f"{self.service.organisation}/{self.service.project}"
 `

` # Try to use the id as it was given, it might be a _self if it starts with self.service.url_resources
            if id.startswith(nexus_path):`

kgforge/specializations/stores/bluebrain_nexus.py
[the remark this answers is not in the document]
crisely09: I meant when you do forge.retrieve(id=id) you need the url_resource value, not the id_without_query, that is why I assign url_resource = id_without_query in the except case.

kgforge/specializations/stores/bluebrain_nexus.py
crisely09: I just tried too, you are correct, the nexus_path was still needed for the cross_bucket case.

Issue #280, opened 2022-11-15
Issues with resource validation


joni-herttuainen:
Hi, I was playing with the KnowledgeGraphForge.validate. There's a few issues.

### Can't pre-validate if there are files
If I call validate on a resource that has files to it, to check if it's valid before registering it, I get:
```
<action> _validate_one
<succeeded> False
<error> ValidationError: resource has lazy actions which need to be executed before
```
if I do:
```
forge.validate(resource, execute_actions_before=True)
<action> _validate_one
<succeeded> False
```
and the validation fails, and if I do not register the resource due to it failing, I have now effectively uploaded a file that will never be used by anything in Nexus.

### Error message is unclear
If I validate a registered resource, in this case `DetailedCircuit`, the output is:
```
<action> _validate_one
<succeeded> False
<error> ValidationError: 
Validation Report
Conforms: False
Results (1):
Constraint Violation in AndConstraintComponent (http://www.w3.org/ns/shacl#AndConstraintComponent):
	Severity: sh:Violation
	Source Shape: this:DetailedCircuitShape
	Focus Node: <https://bbp.epfl.ch/nexus/v1/resources/nse/test2/_/O1.v6a.test>
	Value Node: <https://bbp.epfl.ch/nexus/v1/resources/nse/test2/_/O1.v6a.test>
[...]

PR #281, opened 2022-11-17
Enable debug option for pyschacl validation

crisely09:
In the effort to solve https://github.com/BlueBrain/nexus-forge/issues/280 this option let's us use the `debug` version of the pyschacl validator.


kgforge/core/archetypes/store.py
```diff
@@ -121,15 +121,17 @@ def mapper(self) -> Optional[Callable]:
...
     ) -> None:
         # Replace None by self._register_many to switch to optimized bulk registration.
+        catch_exceptions = False if debug else True
         run(
             self._register_one,
             None,
             data,
             required_synchronized=False,
             execute_actions=True,
+            catch_exceptions=catch_exceptions,
```
MFSY, line 134: I would do:
`catch_exceptions= not debug`

crisely09:
I haven't implemented yet the option to delete distributions and deprecate files when registration fails. I would rather put that in a separate PR.

kgforge/specializations/stores/bluebrain_nexus.py
```diff
@@ -980,28 +984,44 @@ def _initialize_service(
...
+                        messages.append(f"Reason: {detail['resultMessage']} "\
+                                        f"for the shape: {detail['sourceShape']}")
+                elif isinstance(the_details, dict):
+                    messages.append(f"Reason: {detail['resultMessage']} "\
+                                     f"for the shape: {detail['sourceShape']}")
+                else:
+                    messages.append(str(the_details))
+            else:
+                messages.append(str(result))
+    return ". ".join(messages)
```
joni-herttuainen, line 1024: Is it an absolute must to join these? Or, better yet, does the result dictionaries have to be converted to `str`?

crisely09: This is a choice, would you like to have the full schacl error result? Maybe we can give it back when the debug option is true. However, if the intention is to make a nicer Error log, we can add it here, I don't know, what do you think?

joni-herttuainen: No, no. Not the full shacl output, absolutely not. It's just for flexibility of further handling the output.
For example, if the dict you gather above is kept as a dict, it would allow to do something like:
```python
try:
    reason = "Flux Capacitor is broken"
    troubleshooting = {"attention": "Disconnect capacitor drive before opening", "safety": "Shield eyes from light", "actions": "do magic"}
    raise ValueError(reason, troubleshooting)
except ValueError as e:
    print(f"ERROR: {e.args[0]}")
    print("\nTroubleshooting Tips:")
    for k,v in e.args[1].items():
        print(f"\t{k.ljust(10)}: {v}")
```
OUTPUT:
```
ERROR: Flux Capacitor is broken

Troubleshooting Tips:
        attention : Disconnect capacitor drive before opening
        safety    : Shield eyes from light
        actions   : do magic
```
or we can just leave it as it is:
```python
raise ValueError(reason, troubleshooting)
```
OUTPUT:
```
# <traceback>
ValueError: ('Flux Capacitor is broken', {'attention': 'Disconnect capacitor drive before opening', 'safety': 'Shield eyes from light', 'actions': 'do magic'})
```

joni-herttuainen: On nexus-forge side, you would still able to format the messages any way you want in `execution.py`. What do you think?

crisely09: Ah ok ok, I understand now your point, yes, I like it, I will change the printing to use the full dict

joni-herttuainen:
I was testing the newest changes, and we're discussing about it in [NSETM-2067](https://bbpteam.epfl.ch/project/issues/browse/NSETM-2067). 

Instead of opening a new ticket, I'll relay a comment from @jdcourcol, before this is pushed to master:
> This is not great. "RegistrationError" does not bring any information when you are "Registering". The client knows that he is registering. It should be a "ValidationError" exception instead. And then having the information as text is unusable for programmatic access.

kgforge/specializations/models/rdf_model.py
```diff
@@ -107,13 +107,16 @@ def schema_id(self, type: str) -> str:
         except KeyError:
             raise ValueError("type not found")
 
-    def validate(self, data: Union[Resource, List[Resource]], execute_actions_before: bool, type_: str) -> None:
+    def validate(self, data: Union[Resource, List[Resource]], execute_actions_before: bool, type_: str,
+                 debug: bool) -> None:
```
ssssarah, line 111: Very small detail, but the Model class's signature for this method has debug with default value False, and this one doesn't. It creates a bit of a mismatch (IDE complains but no errors)

kgforge/core/archetypes/store.py
```diff
@@ -294,9 +296,10 @@ def _download_one(
     # CR[U]D.
 
     def update(
-        self, data: Union[Resource, List[Resource]], schema_id: Optional[str]
+        self, data: Union[Resource, List[Resource]], schema_id: Optional[str], debug: bool = False
     ) -> None:
         # Replace None by self._update_many to switch to optimized bulk update.
+        catch_exceptions = False if debug else True
```
ssssarah, line 302: In other places, when calling run, you've directly used catch_exceptions=not debug. any reason for the change in style? (in this file you did it also for register, but then for update, tag, deprecate and freeze, the style changed)

ssssarah (commented):
I have yet to run things to see what happens

kgforge/core/archetypes/store.py
crisely09: Ah yes, I believe I initially used this style, and then I was suggested to change it all, but seems I missed many cases

ssssarah: I just saw the suggestion

kgforge/specializations/models/rdf_model.py
crisely09: Changed.

Issue #282, opened 2022-12-06
Download method only uses short url version


crisely09:
With the latest change on the distributions, the values in `contentUrl` (and `_self`) are now expanded to include the org/prod bucket plus the id of the resource. This is not supported by the current `forge.download` method, which detects if the given string is an url, and then cleans the name to use the last part of the url.split('/') to create the "short" version used before.

PR #283, opened 2022-12-12
Use full url to download files.

crisely09:
Related to https://github.com/BlueBrain/nexus-forge/issues/282
The idea is to check if the contentUrl is already expanded and use it as given. 
When the file_id is passed, then do the standard url construction, i.e. endpoint/org/proj/file_id .

### DIFF
--- a/kgforge/specializations/stores/bluebrain_nexus.py
+++ b/kgforge/specializations/stores/bluebrain_nexus.py
@@ -411,7 +411,7 @@ class BlueBrainNexus(Store):
             async with semaphore:
                 params_download = copy.deepcopy(self.service.params.get("download", {}))
                 url_base, org, project = self._prepare_download_one(
-                    url, path, store_metadata, cross_bucket
+                    url, store_metadata, cross_bucket
                 )
                 async with session.get(url_base, params=params_download) as response:
                     try:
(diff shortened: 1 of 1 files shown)


MFSY:
Looks good. Fixed the following errors (introduced in previous merged commits) occurring when downloading:
- Before, _download_many were failing if at least one resource did not have a requested content_type even if other resources in the list had. Now, only resources with the provided content_type are considered when downloading
- Before, Dataset.download() was raising a ValueError when 'follow' was provided because "source" argument has a default value and providing 'source' and 'follow' arguments was not allowed. Now, warnings are raised  (asking to choose one of the two and stating that 'source' will be deprecated) and 'follow' is always used when provided.

MFSY (approved):

crisely09:
I see that you added the solution to the cross_bucket problem here, that is good, two birds in one shot. Maybe we could add an example for downloading using `cross_bucket=True` if it doesn't already exist.

MFSY:
> I see that you added the solution to the cross_bucket problem here, that is good, two birds in one shot. Maybe we could add an example for downloading using `cross_bucket=True` if it doesn't already exist.

It is mentioned both in the Querying and Dataset notebooks.

Issue #262, opened 2022-08-02
KnowledgeGraphForge.sparql(query, rewrite=False) rewrites LIMIT


joni-herttuainen:
__Issue__
When doing a `sparql(..., rewrite=False)` call, the `LIMIT` written in the query string is not reinforced, it needs to be manually fed as a keyword argument. Is this the expected behavior?

I would have expected that `rewrite=False` would guarantee the query string is not manipulated.

This seems to affect at least the latest release and the current version in master.

__Repro__
```python
query = 
"""
# PREFIXES
SELECT ?id WHERE {
    ?id a nsg:DetailedCircuit
} LIMIT 1
"""
forge.sparql(query, debug=True, rewrite=False)
```
Debug output:
```
PREFIXES
SELECT ?id WHERE {
    ?id a nsg:DetailedCircuit
} LIMIT 100
```
To get what I expected:
```python
query = 
"""
# PREFIXES
SELECT ?id WHERE {
    ?id a nsg:DetailedCircuit
}
"""
forge.sparql(query, debug=True, rewrite=False, limit=1)
```

crisely09:
I have a possible (and straightforward) solution: keep the query **as given** when `rewrite=False`, specifically the `limit` and `offset` parameters, and print a warning if any of these parameters is passed to the method, announcing they won't be taken into account because `rewrite` was set to `False`.

crisely09:
Also, it would be good to not have any limit, and just use what is given in the query, or even have an option to not have a limit.

Issue #241, opened 2022-03-21
No way to get unlimited result `forge.elastic`


eugeniashurko:
Currently, when querying resources through `forge.elastic`, there is no way to get all the documents.

For example, when running:

```
resources = forge.elastic("""
     {
        "query": {
              "term" : { "_deprecated": false }
            }
      }
  """, limit=None, debug=True)
```
Only 100 resources are returned. Here is the output with `debug=True`
```
Submitted query: {'query': {'term': {'_deprecated': False}}, 'size': 100, 'from': 0}
```

To be able to retrieve more than 100 resources, some hacks need to be implemented. For example, fixing an unrealistically large limit:
```
resources = forge.elastic("""
     {
        "query": {
              "term" : { "_deprecated": false }
            }
      }
  """, limit=100000, debug=True)
```

Would be great to have a way to get all the documents by specifying `limit=None`, as with `forge.sparql`.

MFSY:
Hi @eugeniashurko ,

There is a [(configured) 10 000 hit limit](https://www.elastic.co/guide/en/elasticsearch/reference/current/paginate-search-results.html#:~:text=By%20default%2C%20you%20cannot%20use,use%20the%20search_after%20parameter%20instead.) when querying ES. It make it difficult to get all docs.

MFSY:
If this is to be addressed, that has to be at index settings level.

eugeniashurko:
I see, it's a good point! But I think from the forge side we should not add this limit `100` for no reason (when limit is specifically None at least), we can just make 'unlimited' queries and then depending on the index, not more than 10k resources will be returned. This behaviour was very unexpected for me and introduced some weird hard-to-debug side-effect.

PR #287, opened 2023-01-17
Removed limit in sparql and elastic when not given. Removed unnecessary arguments

crisely09:
This is a little effort to address two issues:
https://github.com/BlueBrain/nexus-forge/issues/241 and https://github.com/BlueBrain/nexus-forge/issues/262
Basically by not rewriting the offset and limit when they are given as `None`.


kgforge/core/archetypes/store.py
```diff
@@ -369,29 +369,32 @@ def sparql(
...
         )
-        qr = _replace_in_sparql(qr, "LIMIT", limit, 100, r" LIMIT \d+")
-        qr = _replace_in_sparql(qr, "OFFSET", offset, 0, r" OFFSET \d+")
+        use_defaults = params.get("use_defaults", True)
+        if rewrite and use_defaults:
+            qr = _replace_in_sparql(qr, "OFFSET", offset, 0, r" OFFSET \d+")
+            qr = _replace_in_sparql(qr, "LIMIT", limit, 100, r" LIMIT \d+")
         if debug:
             self._debug_query(qr)
         return self._sparql(qr, limit, offset, **params)
```
ssssarah, line 378: sparql calls _sparql with 4 parameters when you've removed the last 2

kgforge/core/archetypes/store.py
```diff
@@ -369,29 +369,32 @@ def sparql(
...
+        self, query: str, debug: bool, limit: int = None, offset: int = None
     ) -> List[Resource]:
         query_dict = json.loads(query)
-        query_dict["size"] = limit if limit else query_dict.get("size", 100)
+        if limit:
+            query_dict["size"] = limit
         query_dict["from"] = offset if offset else query_dict.get("from", 0)
         if debug:
             self._debug_query(query_dict)
         return self._elastic(json.dumps(query_dict), limit, offset)
```
ssssarah, line 395: Same here

ssssarah (changes requested):

kgforge/core/archetypes/store.py
crisely09: True, weird there are no tests failing!! maybe there are no tests for this.

ssssarah (dismissed):

crisely09:
I just added a couple of examples of the use-cases for those parameters to the getting-started notebooks.

crisely09:
@MFSY I put back the elasticsearch default limit, but used a different parameter to `add_context` to a sparql query. I believe this can leave the door open to adding other possible parameters replacement when `rewrite` is set to True.

crisely09:
@ssssarah  could you take another look, please?

kgforge/core/archetypes/store.py
```diff
@@ -378,44 +378,49 @@ def search(
...
         rewrite = params.get("rewrite", True)
         qr = (
             rewrite_sparql(query, self.model_context, self.service.metadata_context)
             if self.model_context is not None and rewrite
             else query
         )
-        qr = _replace_in_sparql(qr, "LIMIT", limit, 100, r" LIMIT \d+")
-        qr = _replace_in_sparql(qr, "OFFSET", offset, 0, r" OFFSET \d+")
+        if limit:
+            qr = _replace_in_sparql(qr, "OFFSET", offset, 0, r" OFFSET \d+")
```
ssssarah, line 398: You swapped the conditions and the corresponding replacing statements

ssssarah: Also this is more of a format thing than a functionality thing, but 100 and 0 are hardcoded in many locations in these files, maybe having a global variable like SPARQL_DEFAULT_LIMIT & ELASTIC_DEFAULT_LIMIT + same for offset (or just DEFAULT_LIMIT and DEFAULT_OFFSET if it's known that both will always have the same defaults) could be better?

kgforge/core/archetypes/store.py
ssssarah: This is still a problem here

kgforge/core/archetypes/store.py
crisely09: yeah, DEFAULT_LIMIT and DEFAULT_OFFSET could work 👍🏽

ssssarah: If a limit is hardcoded, but a limit is specified in the function, the function value should override the hardcoded limit. This works when a number is specified, but if it is specified that the limit should be None, it keeps the hardcoded value. This feels unintentional. I'd think the limit should be removed altogether, no?

ssssarah (approved):

PR #289, opened 2023-01-26
Remove duplicate dependency

heerener:


crisely09 (approved):
Thanks for spotting it!

PR #291, opened 2023-02-17
When downloading files, returns the paths (including filename) where they are downloaded

ssssarah:


crisely09:
@ssssarah this is interesting, but maybe it can be implemented in a different way, to make the return of the list optional.

Also, if you intent to change the stores, then you need also to change the specialized stores in forge, as they are derived from the archetypes.

PR #292, opened 2023-02-22
GROUP BY, ORDER, MINUS Sparql keywords to be ignored by rewriter

ssssarah:


MFSY (approved):

ssssarah:
Any idea why there is there is that "new_base variable used before assignment" error when the tests run?

MFSY:
> Any idea why there is there is that "new_base variable used before assignment" error when the tests run?

 I suspect the new rdflib releases 6.3.X are introducing this error. @ssssarah set the rdflib version to 6.2.0. until we sort this out.

crisely09:
> > Any idea why there is there is that "new_base variable used before assignment" error when the tests run?
> 
> I suspect the new rdflib releases 6.3.X are introducing this error. @ssssarah set the rdflib version to 6.2.0. until we sort this out.

Yes, this is related to my PR https://github.com/BlueBrain/nexus-forge/pull/295 , I also opened an issue on the rdflib side.

PR #304, opened 2023-04-14
Explode filter list when calling SPARQLQueryBuilder from BlueBrainNexus

MFSY:


ssssarah (approved):

crisely09:
In fact I would do the opposite, remove the * from the query builder. What do you think?

crisely09 (approved):

crisely09:
@MFSY should we merge this?

PR #323, opened 2023-07-18
Enabled resolvers mapping files to contains rules calling a forge method

### LINKED ISSUE #316: Support result_resource_mapping with forge calls in Resolver.resolve()
Because of calls like [this](https://github.com/BlueBrain/nexus-forge/blob/master/kgforge/core/archetypes/resolver.py#L99) in Resolver.resolve (i.e no forge object provided), a mapping file with the following content will fail when used as "result_resource_mapping" value of a resolver config:

{
"targetProp":forge.resolve(x.originProp, scope="...", target="...")
}

MFSY:
Fixes #316

### DIFF
--- a/kgforge/core/archetypes/resolver.py
+++ b/kgforge/core/archetypes/resolver.py
@@ -69,7 +69,7 @@ class Resolver(ABC):
 
     def resolve(self, text: Union[str, List[str], Resource], target: str, type: str,
                 strategy: ResolvingStrategy, resolving_context: Any, property_to_resolve: str, merge_inplace_as: str,
-                limit: int, threshold: float) \
+                limit: int, threshold: float, forge: Optional["KnowledgeGraphForge"]) \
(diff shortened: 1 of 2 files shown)


crisely09 (changes requested):
Could you add an example or a test? do we have already a case where the hjson has a forge function?

MFSY:
> Could you add an example or a test? 

This is to match how mappers are invoked in this merged PR https://github.com/BlueBrain/nexus-forge/pull/315/files#diff-c86c44807067dc2a80f4c2bcfa7aa9ec26c580888cc7fbba71032bb14736c73f.

A forge instance should be passed when calling mapper(forge).map()

>do we have already a case where the hjson has a forge function?
There is not yet a configured resolver with a mapping file using a forge method. But this is a matter of time.

crisely09 (approved):

PR #327, opened 2023-08-13
BlueBrainNexusStore: fix tagging schemas

MFSY:
* tagging schemas required a specific schema (https://bluebrain.github.io/nexus/schemas/shacl-20170720.ttl) URI to be present in the request URL
* '_' is present as schema in case of UNCONSTRAINED resource

### DIFF
--- a/kgforge/specializations/stores/nexus/service.py
+++ b/kgforge/specializations/stores/nexus/service.py
@@ -74,6 +74,7 @@ class Service:
     DEFAULT_ES_INDEX_FALLBACK = f"{NEXUS_NAMESPACE_FALLBACK}defaultElasticSearchIndex"
     DEPRECATED_PROPERTY_FALLBACK = f"{NEXUS_NAMESPACE_FALLBACK}deprecated"
     PROJECT_PROPERTY_FALLBACK = f"{NEXUS_NAMESPACE_FALLBACK}project"
+    UNCONSTRAINED_SCHEMA = "https://bluebrain.github.io/nexus/schemas/unconstrained.json"
 
     def __init__(
         self,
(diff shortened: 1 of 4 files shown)


kgforge/specializations/stores/nexus/service.py
```diff
@@ -582,3 +587,27 @@ def to_resource(
         if not hasattr(resource, "id") and kwargs and 'id' in kwargs.keys():
             resource.id = kwargs.get("id")
         return resource
+
+
+def _error_message(error: Union[HTTPError, Dict]) -> str:
```
crisely09, line 592: Why do we make it a string? instead of passing the full dictionary?
Also, I have the feeling that this doesn't work, it does not return the complete error, and that is why I tried to change it in a [this PR ](https://github.com/BlueBrain/nexus-forge/pull/281/files)some time ago.

kgforge/specializations/stores/nexus/service.py
```diff
@@ -449,6 +445,15 @@ async def dispatch_action():
...
         return asyncio.run(dispatch_action())
 
+    def _prepare_tag(self, resource, tag) -> Tuple[str, str, str, str]:
+        schema_id = resource._store_metadata._constrainedBy
+        schema_id = "_" if schema_id == self.UNCONSTRAINED_SCHEMA or schema_id is None else schema_id
+        url = "/".join((self.url_resources, quote_plus(schema_id), quote_plus(resource.id), "tags"))
+        rev = resource._store_metadata._rev
+        params = {"rev":rev}
+        data = {"tag": tag, "rev": rev}
+        return url, data, params, schema_id
```
crisely09, line 455: why returning the schema_id if it is never used?

tests/specializations/stores/test_bluebrain_nexus.py
```diff
@@ -176,6 +177,27 @@ def test_to_resource(nexus_store, registered_building, building_jsonld):
     assert str(result._store_metadata) == str(registered_building._store_metadata)
 
 
+def test_prepare_tag(nexus_store, registered_building):
```
crisely09, line 180: here you are checking the same thing twice, you could use `@pytest.mark.parametrize`

MFSY: Fixed now.

kgforge/specializations/stores/nexus/service.py
MFSY: schema_id is removed

kgforge/specializations/stores/nexus/service.py
MFSY: Let merge this one here and look at the other PR

crisely09: OK

crisely09 (approved):

PR #330, opened 2023-08-25
Raised an exception when a jsonld dict with an @id containing space character is provided to from_jsonld

MFSY:
rdflib.plugins.shared.jsonld.context.Context.resolve(iri) [return ""](https://github.com/RDFLib/rdflib/blob/f34c476ff099b39fe253430cdc25ebc121837e5e/rdflib/plugins/shared/jsonld/context.py#L323) when iri contains space. This leads to silently replacing an iri with "" in   [_remove_ld_keys](https://github.com/BlueBrain/nexus-forge/blob/a936ef152e82afaa1ca0fc019dc4b3f5f79cecae/kgforge/core/conversions/rdf.py#L471).

### DIFF
--- a/tests/core/conversions/test_rdf.py
+++ b/tests/core/conversions/test_rdf.py
@@ -26,7 +26,7 @@ from rdflib.namespace import RDF, Namespace
 
 from kgforge.core import Resource
 from kgforge.core.commons.exceptions import NotSupportedError
-from kgforge.core.conversions.rdf import _merge_jsonld, from_jsonld, as_jsonld, Form, as_graph, from_graph, LD_KEYS
+from kgforge.core.conversions.rdf import _merge_jsonld, _resolve_iri, from_jsonld, as_jsonld, Form, as_graph, from_graph, LD_KEYS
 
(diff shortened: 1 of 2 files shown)


tests/core/conversions/test_rdf.py
```diff
@@ -26,7 +26,7 @@
 
 from kgforge.core import Resource
 from kgforge.core.commons.exceptions import NotSupportedError
-from kgforge.core.conversions.rdf import _merge_jsonld, from_jsonld, as_jsonld, Form, as_graph, from_graph, LD_KEYS
+from kgforge.core.conversions.rdf import _merge_jsonld, _resolve_iri, from_jsonld, as_jsonld, Form, as_graph, from_graph, LD_KEYS
```
crisely09, line 29: this function is nowhere

crisely09: I guess you mean `_remove_ld_keys` ?

MFSY: Added.

crisely09 (approved):

PR #337, opened 2023-09-14
Add extra field to the ontology resolver template and mapping.

crisely09:
The simplest solution to @lecriste's needs.
I did not go for a "costume" creation of the template, as last time when I did it was discarded.


lecriste (commented):
I believe I need only `hasLayerLocationPhenotype`, why are we adding the rest?

MFSY (approved):

PR #338, opened 2023-09-18
Upgrade pyshacl and replace Validator with validate function

crisely09:
Use the latest version of pyshacl, and replace object with newer method.

### DIFF
--- a/kgforge/specializations/models/rdf/directory_service.py
+++ b/kgforge/specializations/models/rdf/directory_service.py
@@ -15,7 +15,7 @@ import os
 from pathlib import Path
 from typing import Dict, Tuple
 
-from pyshacl import Validator
+from pyshacl import validate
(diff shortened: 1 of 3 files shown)


crisely09:
@MFSY this is also ready for review.

crisely09:
This might me causing the error that I was seeing when resolving brain regions:
```
REGION:  hypothalamus
<action> parse
<error> ParserError: Invalid line: ^^http://www.w3.org/2001/XMLSchema#integer .
```

MFSY:
> This might me causing the error that I was seeing when resolving brain regions:
> 
> ```
> REGION:  hypothalamus
> <action> parse
> <error> ParserError: Invalid line: ^^http://www.w3.org/2001/XMLSchema#integer .
> ```

Fixed in https://github.com/BlueBrain/nexus-forge/pull/340

PR #339, opened 2023-09-18
Test PR to see if not using external repo helps the codecov

crisely09:
There is an error in codecov in PR #360, it may be related to the fact that the branch is located in a forked repo.


setup.py
```diff
@@ -51,7 +51,8 @@
         "nest-asyncio>=1.5.1",
         "pyparsing>=2.0.2",
         "owlrl>=5.2.3",
-        "elasticsearch_dsl==7.4.0"
+        "elasticsearch_dsl==7.4.0",
+        "SPARQLWrapper"
```
ssssarah, line 55: Version?

kgforge/specializations/databases/store_database.py
```diff
@@ -0,0 +1,97 @@
...
+        self._check_properties(**config)
+        self.name = config.pop('name')
+        source = config.pop('source')
+        super().__init__(forge, source, **config)
+
+    def _check_properties(self, **info):
+        properties = info.keys()
+        for r in self._REQUIRED:
+            if r not in properties:
+                raise ValueError(f'Missing {r} from the properties to define the DatabasSource')
```
ssssarah, line 56: Little typo (Databas). 
Also if multiples properties are missing, ValueError would fail only for one and then after fixing, would fail for the next. You could check all missing properties first, and then raise the Error with an array of missing properties?

kgforge/core/forge.py
```diff
@@ -936,6 +1011,34 @@ def get_model_context(self):
         """Expose the context used in the model."""
         return self._model.context()
     
+    def create_db_sources(self, all_config: Optional[Dict[str, Dict[str, str]]], 
+                          store_config : Optional[Dict[str, Dict[str, str]]],
+                          model_context: Context
+                          ) -> Union[Database, List[Database]]:
```
ssssarah, line 1017: Return type seems to be Dict[str, Database]

kgforge/core/forge.py
```diff
@@ -936,6 +1011,34 @@ def get_model_context(self):
...
+        names = all_config.keys()
+        dbs = {}
+        for name in names:
```
ssssarah, line 1020: you can also iterate over names with .items()
`for name, config in all_config.items()`

kgforge/core/archetypes/store.py
```diff
@@ -597,3 +600,39 @@ def replace(match: Match) -> str:
     if context.has_vocab():
         pfx = "\n".join([pfx, f"PREFIX : <{context.vocab}>"])
     return f"{pfx}\n{qr}"
+
+
+def resources_from_construct_query(data, context):
```
ssssarah, line 605: fun, we have the same one! https://github.com/BlueBrain/nexus-forge/blob/865ed625b843acd110c9a84be9fbc4fc630a3cba/kgforge/core/commons/sparql_query_builder.py#L198

kgforge/specializations/databases/utils.py
```diff
@@ -0,0 +1,88 @@
...
+            filters = [filter for filter in filters]
+        else:
+            filters = [filters]
+        for filter in filters:
+            if 'type' in filter.path and filter.operator is "__eq__":
+                resource_type = filter.value
+                break
+    return resource_type
+
+def resources_from_results(results):
```
ssssarah, line 43: Similarly to the construct one https://github.com/BlueBrain/nexus-forge/blob/865ed625b843acd110c9a84be9fbc4fc630a3cba/kgforge/core/commons/sparql_query_builder.py#L234

kgforge/core/forge.py
```diff
@@ -936,6 +1011,34 @@ def get_model_context(self):
...
+            config = all_config[name]
+            origin = config['origin']
+            source = config['source']
+            if origin == 'store':
+                # Reuse complete configuration of the store when Nexus is called
+                if source == store_config['name'] == 'BlueBrainNexus':
+                    store_copy = deepcopy(store_config)
+                    with_defaults(config, store_copy,
+                                  "source", "name",
+                                   store_copy.keys())
```
ssssarah, line 1030: Maybe cast store_copy.keys() as a list?

kgforge/core/archetypes/store.py
crisely09: ha! I guess it was a "clear need"

setup.py
crisely09: good question ...

ssssarah:
Having the forge object in the database feels off

setup.py
ssssarah: Or at least the currently used one?

kgforge/specializations/databases/webservice_database.py
```diff
@@ -0,0 +1,196 @@
...
+    def download(self, urls: Union[str, List[str]], 
+                paths: Union[str, List[str]], overwrite: bool = False) -> None:
+        # path: DirPath.
+        """Download files """
+        if isinstance(urls, list):
+            # Check consistancy between urls and paths
+            if not isinstance(paths, list):
+                raise TypeError("Given multiple urls, paths should also be a list.")
+            if len(paths) != len(urls):
+                raise ValueError("Missmatch between urls and paths, they should be the same ammount.")
```
ssssarah, line 121: little typo, double m

ssssarah: You can also put the lengths in the error message

kgforge/specializations/databases/utils.py
```diff
@@ -0,0 +1,88 @@
...
+    """Perform a HTTP request
+    params:
+    -------
+        response_loc : list[str]
+            The nested location of the relevat metadata in the
+            response.
+            Example: NeuroMorpho uses response["_embedded"]["neuronResources"]
+            which should be given as: response_loc = ["_embedded", "neuronResources"]
+    """
+    response_location = params.pop('response_loc', None)
```
ssssarah, line 66: I think it would be clearer to have params as exactly the params being passed to requests.get, and response_loc to be a separate parameter of resources_from_request. You can do the popping if it's necessary before the resources_from_request call

ssssarah: And params can be not keyword arguments but a dict

tests/specializations/databases/test_database.py
ssssarah, line 1: Maybe rebase this to master and add full_path_relative_to_root to your test paths? It fails if my working directory isn't nexus-forge

ssssarah, line 1: Following other tests that use classes, if you use a test as a method inside a class and it's an instance method, make the first parameter the `self` parameter.

kgforge/core/archetypes/database.py
```diff
@@ -0,0 +1,146 @@
...
+            if type_ is None:
+                try:
+                    type_ = resource.type
+                except AttributeError:
+                    mapped_resources.append(resource)
+            elif isinstance(type_, Mapping):
+                mapped_resources.append(self._forge.map(resource_dict, type_))
+            elif type_ in mappings:
+                # type_ is the entity here
+                mapping_class : Mapping = import_class(mappings[type_][0], "mappings")
```
ssssarah, line 82: The `mapping_class` is of type `Type[Mapping]` not `Mapping`

kgforge/core/archetypes/database.py
```diff
@@ -0,0 +1,146 @@
...
+        self.context = self._model.context()
+        if 'model_context' not in config:
+            config['model_context'] = self._model.context()
+        self.source: str = source
+        self.service: Any = self._initialize_service(self.source, **config)
+
+    def __repr__(self) -> str:
+        return repr_class(self)
+
+    def map(self, resources : Union[List[Union[Resource, str]], Union[Resource, str]],
```
ssssarah, line 62: - If I understand correctly, if type_ is not provided, a resource is mapped according to the mapping that is associated with its (the resource’s) type property by model. 
- The if else on the resource type assumes that if it’s not of type Resource, it’s of type Dict, but the typing in map defines it to be a Union (or list of) of Resource and str. Should it be dict instead? 
- If a resource has no type_ and no type is provided, there is a silent error? 
- Similarly there is a series of ifs related to type_ and if neither of the cases that are “valid” are true then it’s silently considering the resource as mapped. Shouldn’t there be some warning? Similar to with the update/register operations, it gives the count of successful operations and failed ones and error messages linked to the failures. 
- (In the last two points I’ve talked about “errors” but really it’s just considering a resource to be mapped when it’s not been. However, the thing that would happen is at least that if the element provided was a dictionary it would be converted into a Resource)

kgforge/core/archetypes/database.py
```diff
@@ -0,0 +1,146 @@
...
+        origin = source_config.pop("origin")
+        if origin == "directory":
+            dirpath = Path(source)
+            return self._service_from_directory(dirpath, **source_config)
+        elif origin == "web_service":
+            source_config['service_context'] = self.context
+            return self._service_from_web_service(source, **source_config)
+        elif origin == "store":
+            store = import_class(source, "stores")
+            if source != 'DemoStore':
```
ssssarah, line 127: Why do you need this check?

ssssarah, line 127: Why do you need this check?

kgforge/core/forge.py
```diff
@@ -47,6 +47,7 @@
 from kgforge.core.wrappings.paths import PathsWrapper, wrap_paths
 from kgforge.specializations.mappers import DictionaryMapper
 from kgforge.specializations.mappings import DictionaryMapping
+from kgforge.specializations.databases import StoreDatabase, WebServiceDatabase
```
ssssarah, line 52: you should declare _db_sources as a property, even if it's not defined on initialisation. So that it's clear all attributes the forge object may have

ssssarah: Or should create_db_sources be called on forge initialisation? It seems like it isn't happening so forge doesn't have db sources

kgforge/core/forge.py
```diff
@@ -936,6 +1011,34 @@ def get_model_context(self):
         """Expose the context used in the model."""
         return self._model.context()
     
+    def create_db_sources(self, all_config: Optional[Dict[str, Dict[str, str]]],
```
ssssarah, line 1014: This doesn't seem to be called anywhere?

kgforge/core/forge.py
crisely09: yes, seems like maybe I changed my mind and returned a dict instead

kgforge/core/archetypes/store.py
crisely09: for this and the next, what would be the cleanest solution? I would like to use your code as it looks more organized, and I agree it fits well in the SPARQLQueryBuilder

ssssarah: Cleanest solution in what sense?

(I've put the resource building methods in SparqlQueryBuilder following @MFSY 's comment, but I am of the opinion that there should be another Class for sparql response parsing, as it has nothing to do with query building. I guess the query builder can be treated as a set of helper functions for sparql, in which case maybe renaming it would be better)

crisely09: Yes, the latter is what it's happening. I would not mind also to have just a class for parsing. 
The question is ... should I keep things as I have, and whenever your branch is implemented I make another PR to use it?

ssssarah: Yes I'm not too sure about that, that's what I had meant on slack that i'd be reluctant to start an implementation for the read only store because it can benefit from work in our 2 branches

ssssarah: I think it's a matter of which is merged first. But I don't think you should try to make the changes on this branch fit the ones on mine, that will happen on the merge to master

Issue #332, opened 2023-09-01
Mapping from a dictionary mapping which cannot find mapping file path doesn't throw any errors


ssssarah:
When using forge.map and providing a Mapping: 

- When initialising the Mapping, if you provide to a mapping a file path and it can't find it, the mapping value will resolve to the mapping file path. See https://github.com/BlueBrain/nexus-forge/blob/02640967dd4034a4a4f5d566f322f2d9fdb9bd42/kgforge/core/archetypes/mapping.py#L64

- When using the Mapper, if it doesn't understand the Mapping value, it simply returns the mapping `self.rules` value, which is an array with as a single value, the "unresolved" file path. See https://github.com/BlueBrain/nexus-forge/blob/02640967dd4034a4a4f5d566f322f2d9fdb9bd42/kgforge/specializations/mappers/dictionaries.py#L62

PR #344, opened 2023-10-12
mapping loading with mapping type specified or different method name

ssssarah:
Adresses https://github.com/BlueBrain/nexus-forge/issues/332

I guess the real issue is that a mapping may be a url, txt or a file path, and if it was neither a url (= couldn't request it) or a filepath (= couldn't load the file), then it was assumed to be a txt mapping. But it could have been intended to be a url but malformed, or a filepath but malformed.
Maybe some validation that whatever string is passed is meant to be a text mapping should be done. I've put a hack of a check that it should start with a {, but I'm not even sure this is true of all text mappings.

Update: in checking the validity of the mapping str provided, it brought out that many current mappings in tests are not valid paths, and the failure to load them was silent and did not make the tests fail

### DIFF
--- a/kgforge/core/archetypes/store.py
+++ b/kgforge/core/archetypes/store.py
@@ -17,11 +17,11 @@ import re
 import time
 from abc import ABC, abstractmethod
 from pathlib import Path
-from typing import Any, Callable, Dict, List, Match, Optional, Tuple, Union
+from typing import Any, Dict, List, Match, Optional, Tuple, Union, Type
 
 from kgforge.core import Resource
+from kgforge.core.archetypes import Mapping, Mapper
 from kgforge.core.commons.attributes import repr_class
-from kgforge.specializations.mappers import DictionaryMapper
 from kgforge.core.commons.context import Context
 from kgforge.core.commons.exceptions import (
     DeprecationError,
@@ -122,7 +122,7 @@ class Store(ABC):
         self.token: Optional[str] = token
         self.versioned_id_template: Optional[str] = versioned_id_template
         loaded = (
-            self.mapping.load(file_resource_mapping) if file_resource_mapping else None
+            self.mapping.load_file(file_resource_mapping) if file_resource_mapping else None
         )
         self.file_mapping: Optional[Any] = loaded
         self.model_context: Optional[Context] = model_context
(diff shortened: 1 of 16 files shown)


kgforge/core/archetypes/store.py
```diff
@@ -122,7 +122,7 @@ def __init__(
         self.token: Optional[str] = token
         self.versioned_id_template: Optional[str] = versioned_id_template
         loaded = (
-            self.mapping.load(file_resource_mapping) if file_resource_mapping else None
+            self.mapping.load_file(file_resource_mapping) if file_resource_mapping else None
```
crisely09, line 125: here, are you sure it is a file? it could be an url ...

crisely09: in fact I am getting the `FileNotFoundError`, for the mapping given in the configuration: `https://raw.githubusercontent.com/BlueBrain/nexus-forge/master/examples/configurations/nexus-store/file-to-resource-mapping.hjson`

crisely09: changing it back to `load` makes things work fine

tests/specializations/mappings/test_mappings.py
```diff
@@ -13,3 +13,43 @@
...
+    with exception:
+        mapping = DictionaryMapping.load(source)
+
+
+@pytest.mark.parametrize(
+    "source, mapping_type, exception",
+    [
+        (mapping_path_valid, MappingType.FILE, does_not_raise()),
+        (mapping_url_valid, MappingType.URL, does_not_raise()),
+        ("i", MappingType.URL, pytest.raises(Exception)),
```
crisely09, line 50: you can add more of the exception cases, to complete the testing (file path being passed as a url, text being fiven as file, url as text, etc.)

kgforge/core/archetypes/store.py
ssssarah: My mistake

tests/specializations/mappings/test_mappings.py
ssssarah: I hadn't yet because there was no usage of these, but sure

ssssarah: I've added some test cases

crisely09 (changes requested):

crisely09 (approved):
Looks good

PR #348, opened 2023-10-16
Refactor services

ssssarah:
Services are being initialised in Resolver, Store and Model but there was no typing and abstraction that defined what these services were. In this PR, there some reorganising of where the services are, what classes they extend, and which classes can be used as a service for the Resolver, Store and Model


crisely09:
This is not a bad idea, but now that we are looking into the services, maybe we also avoid initializing the same store/service multiple times when creating a forge instance, what do you think?

ssssarah:
Agreed but I see this as a first step in improving readability

PR #351, opened 2023-10-24
load ontologies

ssssarah:


tests/data/shacl-model/commons/ontology-1.ttl
ssssarah, line 1: This is the parameter ontology, I assume a dummy ontology file would be more appropriate

kgforge/core/archetypes/model.py
```diff
@@ -110,6 +118,35 @@ def _template(self, type: str, only_required: bool) -> Dict:
...
+            limit=limit,
+            offset=offset,
+            default_limit=DEFAULT_LIMIT,
+            default_offset=DEFAULT_OFFSET,
+            debug=debug
+        )
+
+        return self._sparql(qr)
+
+    def _sparql(self, query: str) -> SPARQLResult:
```
ssssarah, line 144: This is the unprocessed return value and type from the graph object being queried, I'm not sure which format is the most appropriate

kgforge/core/commons/sparql_query_rewriter.py
```diff
@@ -0,0 +1,173 @@
...
+    "group",
+    "by",
+    "order",
+    "minus",
+    "not",
+    "exists"
+]
+
+
+def rewrite_sparql(query: str, context: Context, metadata_context: Context) -> str:
```
MFSY, line 58: This can go to the SPARQLQueryBuilder.

kgforge/core/commons/sparql_query_rewriter.py
```diff
@@ -0,0 +1,173 @@
...
+
+    if is_what_in_query and replace_if_in_query and replace_value:
+        qr = re.sub(f"{search_regex}", replace_value, qr, flags=re.IGNORECASE)
+
+    if not is_what_in_query and replace_value:
+        qr = f"{qr} {replace_value}"
+    return qr
+
+
+def handle_sparql_query(
```
MFSY, line 155: This method's name would benefit from being moire clear. Either all this logic is in the rewrite_sparql methods (after all changing the limit or offset is also rewriting the query) or the limit and offset change is in its own method.

kgforge/core/commons/sparql_query_rewriter.py
```diff
@@ -0,0 +1,173 @@
...
+    if offset:
+        qr = _replace_in_sparql(qr, "OFFSET", offset, default_offset, r" OFFSET \d+")
+
+    if debug:
+        _debug_query(qr)
+
+    return qr
+
+
+def _debug_query(query):
```
MFSY, line 187: debug here has nothing to do with sparql rewriting. I think this can stay where it was

kgforge/specializations/models/demo_model.py
```diff
@@ -96,8 +96,8 @@ def _validate_one(self, resource: Resource, type_: str) -> None:
     # Utils.
 
     @staticmethod
-    def _service_from_directory(dirpath: Path, context_iri: str, **dir_config):
-        return ModelLibrary(dirpath)
+    def _service_from_directory(dir_path: Path, context_iri: Optional[str]):
```
MFSY, line 99: (dir_path) This is changing the Model interface API. Unless really necessary, I would not rename this. Also any change on this method should also be done on the Model interface.

kgforge/specializations/models/rdf/rdf_model_service_from_directory.py
```diff
@@ -73,18 +72,31 @@ def _build_shapes_map(self) -> Dict:
...
                     }
                 }
-            } ORDER BY ?type"""
+            } ORDER BY ?type
+        """
         res = self._graph.query(query)
-        return {row["type"]: row["shape"] for row in res}
 
+        class_being_shaped_id_to_shape_uri: Dict[str, URIRef] = {
+            row["type"]: URIRef(row["shape"])
```
MFSY, line 80: Why are propagating URIRef instead of str ?

kgforge/specializations/models/rdf/rdf_model_service_from_directory.py
```diff
@@ -73,18 +72,31 @@ def _build_shapes_map(self) -> Dict:
                         ?shape a rdfs:Class
                     }
                 }
-            } ORDER BY ?type"""
+            } ORDER BY ?type
+        """
         res = self._graph.query(query)
-        return {row["type"]: row["shape"] for row in res}
 
+        class_being_shaped_id_to_shape_uri: Dict[str, URIRef] = {
```
MFSY, line 79: What class_being_shaped_id_to_shape_uri means ?

kgforge/specializations/models/rdf/rdf_model_service_from_directory.py
```diff
@@ -73,18 +72,31 @@ def _build_shapes_map(self) -> Dict:
...
+        # FIXME should return the file path where the schema is in
+        schema_to_file = dict(
+            (e, "")  # TODO file source
+            for e in class_being_shaped_id_to_shape_uri.values()
+        )
+
+        return schema_to_file, class_being_shaped_id_to_shape_uri
+
+
+def load_rdf_files_into_graph(path: Path, memory_graph: Graph) -> Graph:
```
MFSY, line 93: I would not recommend to pass an empty graph and mutate it inside this method. What are you gaining by doing so compare to initialising, building the graph and returning it from this method ?

kgforge/specializations/models/rdf/rdf_model_service_from_store.py
```diff
@@ -59,24 +61,22 @@ def materialize(self, iri: URIRef) -> NodeProperties:
...
+
+        return self._context_cache[iri]
 
     def generate_context(self) -> Dict:
-        for v in self._shapes_to_resources.values():
-            self._load_shape(v)
-        # reloads the shapes graph
-        self._sg = ShapesGraphWrapper(self._graph)
+        for v in self.schema_to_source.values():
+            self._load_shape_and_reload_shapes_graph(v)
```
MFSY, line 75: I think one thing doing one thing is better than _load_shape_and_reload_shapes_graph

kgforge/specializations/models/rdf/rdf_model_service_from_store.py
```diff
@@ -91,22 +91,24 @@ def _build_shapes_map(self) -> Dict:
...
+        schema_to_resource: Dict[URIRef, URIRef] = dict()
+
         while count == limit:
             resources = self.context_store.sparql(query, debug=False, limit=limit, offset=offset)
             for r in resources:
                 shape_uri = URIRef(r.shape)
-                class_to_shapes[r.type] = shape_uri
-                shape_resource[shape_uri] = URIRef(r.resource_id)
+                class_being_shaped_id_to_shape_uri[r.type] = shape_uri
+                schema_to_resource[shape_uri] = URIRef(r.resource_id)
```
MFSY, line 107: Actually it is shape to resource and schema to resource. To avoid confusion with what we call schema elsewhere

kgforge/specializations/models/rdf/rdf_model_service_from_store.py
```diff
@@ -91,22 +91,24 @@ def _build_shapes_map(self) -> Dict:
...
+        class_being_shaped_id_to_shape_uri = dict()
+        schema_to_resource: Dict[URIRef, URIRef] = dict()
+
         while count == limit:
             resources = self.context_store.sparql(query, debug=False, limit=limit, offset=offset)
             for r in resources:
                 shape_uri = URIRef(r.shape)
-                class_to_shapes[r.type] = shape_uri
-                shape_resource[shape_uri] = URIRef(r.resource_id)
+                class_being_shaped_id_to_shape_uri[r.type] = shape_uri
```
MFSY, line 106: Looks like a very complicated name => class_being_shaped_id_to_shape_uri

kgforge/specializations/models/rdf/rdf_model_service_from_store.py
```diff
@@ -115,24 +117,27 @@ def recursive_resolve(self, context: Union[Dict, List, str]) -> Dict:
...
                 except ValueError as e:
                     raise e
+
             document.update(self.recursive_resolve(doc))
         elif isinstance(context, dict):
             document.update(context)
         return document
 
-    def _load_shape(self, resource_id):
+    def _load_shape(self, resource_id: URIRef):
```
MFSY, line 140: Why resource_id is turned into URIRef here ? store.retrieve expects an str

kgforge/specializations/models/rdf/rdf_model_service_from_store.py
```diff
@@ -151,12 +159,16 @@ def _load_shape(self, resource_id):
                     for dependency in shape.imports:
                         self._load_shape(self.context.expand(dependency))
 
-    def _type_shape(self, iri: URIRef):
+    def _load_and_get_type_shape(self, iri: URIRef) -> ShapeWrapper:
```
MFSY, line 162: The actual return type is Shape.

kgforge/specializations/models/rdf_model.py
```diff
@@ -87,27 +89,30 @@ def _generate_context(self) -> Context:
...
+            raise ValueError(f"Type {type} not found")
+        return self.service.types_to_shapes[type]
 
-    def schema_id(self, type: str) -> str:
-        try:
-            shape_iri = self.service.types_to_shapes[type]
-            return str(self.service.schema_source_id(shape_iri))
-        except KeyError:
-            raise ValueError("type not found")
+    def schema_id(self, type: str) -> URIRef:
```
MFSY, line 108: Why URIRef ?

tests/data/shacl-model/commons/ontology-1.ttl
```diff
@@ -0,0 +1,375 @@
...
+@prefix ns1: <http://purl.org/vocab/vann/> .
+@prefix nsg: <https://neuroshapes.org/> .
+@prefix owl: <http://www.w3.org/2002/07/owl#> .
+@prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
+@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .
+@prefix prov: <http://www.w3.org/ns/prov#> .
+@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
+@prefix skos: <http://www.w3.org/2004/02/skos/core#> .
+@prefix parms: <https://bbp.epfl.ch/ontologies/core/parameters/> .
+@prefix schema: <http://schema.org/> .
```
MFSY, line 11: Why do you need this ontology ?

MFSY (changes requested):

tests/data/shacl-model/commons/ontology-1.ttl
ssssarah: Just so that I can write a test to test a query against an ontology

kgforge/core/commons/sparql_query_rewriter.py
ssssarah: Just pushed some changes to put them all in the builder

kgforge/core/commons/sparql_query_rewriter.py
ssssarah: I've put it in the query builder, isn't the purpose of the debug to actually make sure the rewriting went well?

kgforge/core/commons/sparql_query_rewriter.py
ssssarah: the elastic search querying has a similar logic so application of limit and offset can be a method of its own defined in the query builder

kgforge/specializations/models/demo_model.py
ssssarah: It's not really necessary it's just to have consistent naming. What do you mean by the model interface API?

kgforge/specializations/models/rdf/rdf_model_service_from_store.py
ssssarah: Yes I wasn't paying much attention to my usage of schema/shape but i can use shape exclusively

kgforge/specializations/models/rdf/rdf_model_service_from_directory.py
ssssarah: Initially I had called the load_rdf_files_into_graph multiple times, for multiple paths (ontology vs schema). This is no longer the case, but it was in order to reuse the same graph object and load into it files from different locations

kgforge/specializations/models/rdf/rdf_model_service_from_directory.py
ssssarah: I've renamed it to class_to_shape. But it was supposed to be:
key: id of the class that is the target of the shape
value: uri of the shape

kgforge/specializations/models/rdf/rdf_model_service_from_store.py
ssssarah: I've renamed it back to class_to_shape. But it was supposed to be:
key: id of the class that is the target of the shape
value: uri of the shape

Initially class_to_shape wasn't sufficient for me to understand what was in it so I made it a bit more verbose

kgforge/specializations/models/rdf_model.py
ssssarah: https://github.com/BlueBrain/nexus-forge/blob/master/kgforge/specializations/models/rdf/store_service.py#L103 it's a consequence of this line. The dictionary value was set to be a URIRef for class_to_shapes and this is used by _build_types_to_shapes to create types_to_shapes and get_shape_from_type gets it from it. So I just added types

kgforge/specializations/models/rdf/rdf_model_service_from_store.py
ssssarah: My mistake, just changed it

kgforge/specializations/models/rdf/rdf_model_service_from_directory.py
ssssarah: https://github.com/BlueBrain/nexus-forge/blob/master/kgforge/specializations/models/rdf/store_service.py#L103 it's a consequence of this line. The dictionary value was set to be a URIRef for class_to_shapes and this is used by _build_types_to_shapes to create types_to_shapes and get_shape_from_type gets it from it. So I just added types

kgforge/specializations/models/rdf/rdf_model_service_from_store.py
ssssarah: The _graph and the _shapes_graph are supposed to stay consistent so it's better to always _load_shape and put it in the shapes_graph automatically, rather than expect the person calling _load_shape to keep them consistent

kgforge/specializations/models/rdf/rdf_model_service.py
```diff
@@ -11,137 +11,47 @@
...
-    def schema_source_id(self, schema_iri: str) -> str:
-        # POLICY Should return the id of the resource containing the schema
-        raise NotImplementedError()
+    def shape_source(self, schema_iri: URIRef) -> str:
+        return self.shape_to_source[schema_iri]
+
+    def sparql(self, query: str) -> List[Resource]:
+        e = self._graph.query(query)
+        results = json.loads(e.serialize(format="json"))
+        return SPARQLQueryBuilder.build_resource_from_select_query(results["results"]["bindings"])
```
ssssarah, line 54: Should there be support for other types of queries like in store (ex: Construct queries) ?

kgforge/specializations/models/rdf/rdf_model_service_from_directory.py
crisely09: does it mean that all the files inside a directory will be used to populate the graph?

ssssarah: Yes.
In my most recent commit, I get rid of this method completely, so that I can implement the method that was necessary for this [test](https://github.com/BlueBrain/nexus-forge/blob/ab22a830b7d4ed0fc9e01392fe6ff88504a3594b/tests/specializations/models/test_rdf_model.py#L99) to be valid.
It also follows the fix me [here](https://github.com/BlueBrain/nexus-forge/blob/ab22a830b7d4ed0fc9e01392fe6ff88504a3594b/kgforge/specializations/models/rdf/directory_service.py#L35)


I think it's worth looking at the [commit](https://github.com/BlueBrain/nexus-forge/pull/351/commits/8a5dfee3b69abb30c2f16c0c71364c92104cfba3) separately since it's targeting only this feature. I've put an explanation of why I made these changes as a comment in this test.

crisely09 (approved):

tests/data/shacl-model/commons/ontology-1.ttl
crisely09: Yes, you could keep just a part of the file, not all of it

ssssarah:
Hasn't been updated in a while, out of sync, to refresh

PR #352, opened 2023-10-26
pycodestyle

ssssarah:
Inspired by https://github.com/BlueBrain/nexus-forge/pull/237 , but on the more recent codebase.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,5 +1,5 @@
 [MESSAGES CONTROL]
-disable=C0115,C0114,C0116,W0511,W0123,W0246,W0212
+disable=C0115,C0114,C0116
 
 [FORMAT]
 # Maximum number of characters on a single line.
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,9 +1,9 @@
 [MESSAGES CONTROL]
-disable=bad-continuation,fixme,invalid-name,len-as-condition,no-else-return
+disable=C0115,C0114,C0116
 
 [FORMAT]
 # Maximum number of characters on a single line.
-max-line-length=100
+max-line-length=200
 
 [DESIGN]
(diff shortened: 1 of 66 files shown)


.pylintrc
ssssarah, line 1: Disabled docstring warnings because there are many of them, but in the future these should be re-enabled and docstrings should be written

MFSY:
> Inspired by #237 , but on the more recent codebase. Running `tox -e lint` led to many warnings. This is just addressing some of them. Maybe running the command should be done as job systematically (as proposed in the previously mentioned MR).

Please add the job.

ssssarah:
> > Inspired by #237 , but on the more recent codebase. Running `tox -e lint` led to many warnings. This is just addressing some of them. Maybe running the command should be done as job systematically (as proposed in the previously mentioned MR).
> 
> Please add the job.

I've just added the job in the github workflows, but it only runs pycodestyle because I've fixed all pycodestyle errors, but I haven't fixed all pylint errors, so if added the pylint one it would fail

.pylintrc
ssssarah: - W0511: TODO-FIXME
- W0123: use of eval
- W0212: Access to protected member
- W0246: redefinition of constructor in subclass, but only calling super (nothing else added)

ssssarah: (I added some ignoring in order to have a better view of the warnings I can easily address but I don't actually believe these should be ignored)

crisely09:
It's fine with me. I will approve but let @MFSY have the last word.

crisely09 (approved):

ssssarah:
Although I don't think it helps much, I've re-enabled pylint and added a threshold under which the job fails. The score should be above 8.5/10

MFSY (approved):

PR #353, opened 2023-10-30
Archetypes for external datasets

crisely09:
- Add ReadStore
- Make Store inherit ReadStore
- Add ExternalDataset as ReadStore + other methods (types)

Missing:
- forge methods to access external datasets
- specializations
- tests


kgforge/core/archetypes/read_store.py
```diff
@@ -0,0 +1,222 @@
...
+            store_metadata: Optional[List[DictWrapper]],
+            cross_bucket: bool,
+            content_type: str,
+            buckets: List[str]
+    ) -> None:
+        # paths: List[FilePath].
+        # Bulk downloading could be optimized by overriding this method in the specialization.
+        # POLICY Should follow self._download_one() policies.
+        for url, path, store_m in zip(urls, paths, store_metadata):
+            self._download_one(url, path, store_m, cross_bucket, content_type)
```
ssssarah, line 148: Probably it a problem that was already there but the buckets argument is not used, and the _download_one call is missing a bucket parameter

crisely09: good catch

kgforge/specializations/stores/demo_store.py
```diff
@@ -131,6 +132,12 @@ def search(self, resolvers: Optional[List[Resolver]], *filters, **params) -> Lis
         records = self.service.find(conditions)
         return [_to_resource(x) for x in records]
 
+    def _sparql():
```
ssssarah, line 135: should have the self parameter, same for elastic

crisely09: oh right, keep missing those

kgforge/core/archetypes/external_dataset.py
```diff
@@ -0,0 +1,133 @@
...
+        else:
+            # Try to find the type of the resources within the filters
+            resource_type = type_from_filters(*filters)
+            return self.map(unmapped_resources, type_=resource_type)
+
+    @abstractmethod
+    def _search(self):
+        ...
+
+    def sparql(self, query: str, debug: bool = False, limit: Optional[int] = None,
```
ssssarah, line 54: doesn't have the same signature as the readstore's sparql method. I think you add parameters to the readstore's sparql method, and it should be the same for the external dataset's and the store

kgforge/core/archetypes/external_dataset.py
```diff
@@ -0,0 +1,133 @@
...
+        keep_original = params.pop('keep_original', True)
+        unmapped_resources = self._search(resolvers, *filters, **params)
+        if isinstance(self.service, BlueBrainNexus) or keep_original:
+            return unmapped_resources
+        else:
+            # Try to find the type of the resources within the filters
+            resource_type = type_from_filters(*filters)
+            return self.map(unmapped_resources, type_=resource_type)
+
+    @abstractmethod
```
ssssarah, line 50: abstractmethod is not imported

kgforge/core/archetypes/external_dataset.py
```diff
@@ -0,0 +1,133 @@
...
+    def search(self, resolvers, *filters, **params):
+        """Search within the database.
+
+        :param keep_original: bool
+        """
+        keep_original = params.pop('keep_original', True)
+        unmapped_resources = self._search(resolvers, *filters, **params)
+        if isinstance(self.service, BlueBrainNexus) or keep_original:
+            return unmapped_resources
+        else:
```
ssssarah, line 45: unnecessary else, linting would complain

kgforge/core/archetypes/external_dataset.py
crisely09: ah, let me check

kgforge/core/archetypes/store.py
```diff
@@ -443,11 +303,9 @@ def sparql(
             self._debug_query(qr)
         return self._sparql(qr)
 
-    def _sparql(self, query: str) -> List[Resource]:
-        # POLICY Should notify of failures with exception QueryingError including a message.
-        # POLICY Resource _store_metadata should not be set (default is None).
-        # POLICY Resource _synchronized should not be set (default is False).
-        not_supported()
+    @abstractmethod
+    def _sparql():
```
ssssarah, line 307: should have a self argument, and a query: str argument if you expect that this _sparql should be called by specifying only one str parameter (which is how it's called in sparql(..) )

kgforge/core/archetypes/external_dataset.py
crisely09: I don't understand how VSCode is not complaining, or maybe I am not seeing it

kgforge/core/archetypes/store.py
ssssarah: I guess it should be the way it was before except it's now abstract

kgforge/core/archetypes/read_store.py
```diff
@@ -0,0 +1,223 @@
...
+
+    @staticmethod
+    def _debug_query(query) -> None:
+        if isinstance(query, Dict):
+            print("Submitted query:", query)
+        else:
+            print(*["Submitted query:", *query.splitlines()], sep="\n   ")
+        print()
+
+    def rewrite_uri(self, uri: str, context: Context, **kwargs) -> str:
```
ssssarah, line 217: Should be abstract or the body should have not_supported()

kgforge/core/archetypes/store.py
ssssarah, line 1: sparql signature doesn't fit the read store's signature due to the default value of the limit and offset. Either the read store says the default is None, and so does the sparql method in the store (which wouldn't be great because None is a conscious value choice and not the absence of a value choice), or it says that they all share the default limit and offset

kgforge/core/archetypes/external_dataset.py
```diff
@@ -0,0 +1,134 @@
...
+    def types(self):
+        # TODO: add other datatypes used, for instance, inside the mappings
+        return list(self.model.mappings(self.model.source, False).keys())
+
+    def search(self, resolvers, *filters, **params):
+        """Search within the database.
+
+        :param keep_original: bool
+        """
+        keep_original = params.pop('keep_original', True)
```
MFSY, line 44: I would recommend to use 'map:True|False' (or transform_result) instead of 'keep_original:True|False'.

kgforge/core/archetypes/external_dataset.py
```diff
@@ -0,0 +1,134 @@
...
+        unmapped_resources = self._search(resolvers, *filters, **params)
+        if isinstance(self.service, BlueBrainNexus) or keep_original:
+            return unmapped_resources
+        # Try to find the type of the resources within the filters
+        resource_type = type_from_filters(*filters)
+        return self.map(unmapped_resources, type_=resource_type)
```
MFSY, line 50: Is map() defined ? I did not see it in the is PR.

kgforge/core/archetypes/external_dataset.py
```diff
@@ -0,0 +1,134 @@
...
+        return list(self.model.mappings(self.model.source, False).keys())
+
+    def search(self, resolvers, *filters, **params):
+        """Search within the database.
+
+        :param keep_original: bool
+        """
+        keep_original = params.pop('keep_original', True)
+        unmapped_resources = self._search(resolvers, *filters, **params)
+        if isinstance(self.service, BlueBrainNexus) or keep_original:
```
MFSY, line 46: Why checking if BlueBrainNexus (a specific Store) is the service in this class ? Shouldn't 'keep_original==True' be the only condition for returning unmapped resources. Can BlueBrainNexus be a service (i.e return value of init_service) ?

kgforge/core/archetypes/external_dataset.py
```diff
@@ -0,0 +1,134 @@
...
+        else:
+            filters = [filters]
+        for filter in filters:
+            if 'type' in filter.path and filter.operator is "__eq__":
+                resource_type = filter.value
+                break
+    return resource_type
+
+
+def resources_from_request(url, headers, **params):
```
MFSY, line 89: I did not see being used. I recommend to add it later once it is used. Meanwhile, I think this can go to utils. It seems yet not quite generic ('response_loc'). One way to handle this could be to make the request and then call a callback method for result handling. The caller provide the callback method

MFSY (changes requested):

kgforge/core/archetypes/__init__.py
```diff
@@ -17,5 +17,6 @@
 from .mapping import Mapping
 from .mapper import Mapper
 from .store import Store
+from .read_store import ReadStore
```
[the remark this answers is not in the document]
crisely09: OK

kgforge/core/archetypes/external_dataset.py
```diff
@@ -0,0 +1,134 @@
...
+from typing import Optional, Union, List
+from kgforge.core import Resource
+from kgforge.core.archetypes.read_store import ReadStore
+from kgforge.core.archetypes.model import Model
+from kgforge.core.commons.context import Context
+from kgforge.core.commons.exceptions import QueryingError
+from kgforge.specializations.stores.bluebrain_nexus import BlueBrainNexus
+
+
+class ExternalDataset(ReadStore):
```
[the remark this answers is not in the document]
crisely09: OK, I would make it DatasetStore then

kgforge/core/forge.py
```diff
@@ -227,6 +227,19 @@ def __init__(self, configuration: Union[str, Dict], **kwargs) -> None:
         # Store.
         store_config.update(model_context=self._model.context())
         store_name = store_config.pop("name")
+        store_model_config = store_config.pop("model", None)
+        if store_model_config:
+            store_model_name = store_model_config.pop("name")
+            if store_model_config:
+                if store_model_name != model_name:
+                    # Same model, different config
```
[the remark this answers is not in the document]
crisely09: you mean that we expect either to use it or not using any model?, is there no case where one would like to have the model of the store to be different to the one initialized with forge?

kgforge/core/forge.py
```diff
@@ -227,6 +227,19 @@ def __init__(self, configuration: Union[str, Dict], **kwargs) -> None:
...
+                    store_model = import_class(store_model_name, "models")
+                    store_config['model'] = store_model(**store_model_config)
+                else:
+                    # Same model, same config
+                    store_config['model'] = self._model
+        else:
+            raise ValueError(f'Missing model configuration for store {store_name}')
```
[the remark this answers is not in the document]
crisely09: At this moment yes, but that can be removed.

tests/conftest.py
```diff
@@ -175,6 +175,7 @@ def forge():
         },
         "Store": {
             "name": "DemoStore",
+            "model": {"name": "DemoModel"}
```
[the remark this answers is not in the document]
crisely09: Correct, I haven't yet just to make sure first we agree on it. Will add it now.

kgforge/core/archetypes/read_store.py
```diff
@@ -0,0 +1,227 @@
...
+        # POLICY Resource _store_metadata should be set using wrappers.dict.wrap_dict().
+        # POLICY Resource _synchronized should be set to True.
+        # TODO These two operations might be abstracted here when other stores will be implemented.
+        ...
+
+    @abstractmethod
+    def sparql(self, query: str, debug: bool, limit: int = DEFAULT_LIMIT,
+               offset: int = DEFAULT_OFFSET,
+               **params) -> Optional[Union[List[Resource], Resource]]:
+        ...
```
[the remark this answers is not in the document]
crisely09: I think I removed some that didn't hold anymore, I can check again what was removed.

kgforge/core/archetypes/store.py
crisely09: changed

tests/conftest.py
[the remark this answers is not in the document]
crisely09: added the changes in the config files

kgforge/core/archetypes/external_dataset.py
crisely09: sorry, I was missing it, now it is there

kgforge/core/archetypes/external_dataset.py
crisely09: yeah, that is removed, it doesn't depend on the service

crisely09:
@MFSY you requested changes, this needs your approval anyway

kgforge/core/archetypes/read_only_store.py
```diff
@@ -0,0 +1,227 @@
...
+        # POLICY Resource _synchronized should be set to True.
+        # TODO These two operations might be abstracted here when other stores will be implemented.
+        ...
+
+    @abstractmethod
+    def sparql(self, query: str, debug: bool, limit: int = DEFAULT_LIMIT,
+               offset: int = DEFAULT_OFFSET,
+               **params) -> Optional[Union[List[Resource], Resource]]:
+        ...
+
```
crisely09, line 198: add elastic method

kgforge/core/archetypes/read_only_store.py
```diff
@@ -0,0 +1,227 @@
...
+                                content_type, buckets)
+        elif len(download_urls) == 1:
+            self._download_one(download_urls[0], filepaths[0], download_store_metadata[0],
+                               cross_bucket, content_type, buckets[0])
+        else:
+            raise DownloadingError(
+                f"No resource with content_type {content_type} was found when following the resource path '{follow}'."
+            )
+
+    def _download_many(
```
crisely09, line 138: do download instead of one or many (same for the other methods)

crisely09:
Merged in the SPARQL store specialization

PR #354, opened 2023-10-30
return None if pretty in model archetype

ssssarah:
Some functions in the model archetype were returning an optional string, but for some cases in if statements there were no return statements

### DIFF
--- a/kgforge/core/archetypes/model.py
+++ b/kgforge/core/archetypes/model.py
@@ -52,14 +52,15 @@ class Model(ABC):
     # Vocabulary.
 
     def prefixes(self, pretty: bool) -> Optional[Dict[str, str]]:
-        prefixes = sorted(self._prefixes().items())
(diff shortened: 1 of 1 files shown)


crisely09:
Oh, but I think they should return the mappings if they exist, not `None`

ssssarah:
Yes but that changes the behavior of things. I just made this for code quality

crisely09:
OK, now they are consistent, they are either returned as strings or as dict, not both, correct?

ssssarah:
The way I understand it is that if you specify pretty=True then you want it to be in the standard output stream and nothing is being returned, and it it's false you want it as a dict. But it's never returned as a str. I'm not against it but my goal with the MR was just to be explicit about the returning of nothing, not to change behavior

crisely09 (approved):
Fine with me.

PR #355, opened 2023-10-30
More linting

ssssarah:
- rm import from kgforge.commons.archetypes and kgforge.core and target specific files to avoid cyclic dependencies
- When catching an exception and re-raising a different one, raise from the one that was caught to keep the stack trace and facilitate debugging
- not_supported to return exception to be raised
- not_supported raised in specialisations, not core. In core, methods defined as abstract. Motivation: when implementing one of the archetypes, it's simpler to extend it and to have your IDE generate all abstract methods than it is to have them "implemented for you as not implemented" and having to go to the archetype definition to retrieve the method signatures to override. Moreover, it's better to have a complete visual of unimplemented functionality in one's specialisation, rather than to have to go back to the archetype to see methods that have yet to be implemented
- rm redefinition of constructor in subclasses if no additional logic is added
- rm raise Exception, raise more specific exceptions
- Raising linting threshold to 9.0


kgforge/specializations/stores/demo_store.py
```diff
@@ -139,6 +139,44 @@ def _initialize_service(
     ):
         return StoreLibrary()
 
+    def _register_many(self, resources: List[Resource], schema_id: str) -> None:
```
crisely09, line 142: wait, where these methods not here?

tox.ini
```diff
@@ -24,7 +24,7 @@ deps =
     pylint
 commands =
     pycodestyle {[base]name}
-    pylint -j2 {[base]name} --fail-under=8.5
+    pylint -j2 {[base]name} --fail-under=9.0
```
crisely09, line 27: nice, bit by bit

kgforge/specializations/stores/demo_store.py
crisely09: never mind, they were part of the `not_supported()` methods in the archetype

crisely09:
These are small but wide changes, can we merge this after the archetypes change?

tox.ini
ssssarah: hehe

ssssarah:
> These are small but wide changes, can we merge this after the archetypes change?

Yeah of course, archetypes > everything else

kgforge/specializations/stores/demo_store.py
ssssarah: And that is one of the reasons why I find it best to have the not_supported() in the implementation, because you're not even aware of all that can be implemented otherwise

kgforge/specializations/resolvers/demo_resolver.py
```diff
@@ -63,7 +59,7 @@ def _resolve(self, text: Union[str, List[str]], target: Optional[str], type: Opt
                 )
             except StopIteration:
                 return None
-        elif strategy == ResolvingStrategy.EXACT_CASEINSENSITIVE_MATCH:
+        elif strategy == ResolvingStrategy.EXACT_CASE_INSENSITIVE_MATCH:
```
crisely09, line 62: if you are changing this, you should also update the notebook in use-cases about resolving

ssssarah: It's in both, I just pushed the change

tox.ini
crisely09, line 27: nice

crisely09 (approved):
Good with me.

PR #356, opened 2023-11-06
Increase evaluation path limit

ssssarah:
A bit ugly, but addresses https://bbpteam.epfl.ch/project/issues/browse/DKE-1238

Overrides the pyshacl Shape's [validation method](https://github.com/RDFLib/pySHACL/blob/master/pyshacl/shape.py#L421) and increases the path length that triggers the Evaluation path too deep error

kgforge/specializations/models/rdf/shape_validate.py `shape_validate` is equivalent to its [existing implementation](https://github.com/RDFLib/pySHACL/blob/master/pyshacl/shape.py#L421) except that on this [line](https://github.com/RDFLib/pySHACL/blob/52e239c41b88497e9c9f5fd8ae0ae567e59885b8/pyshacl/shape.py#L467) , the value 30 is replaced by 40


lecriste:
I see only new lines, no changed or removed lines. How is the previous pyshacl Shape method not executed?

lecriste (commented):
I see only new lines, no changed or removed lines. How is the previous pyshacl Shape method not executed now?

ssssarah:
> I see only new lines, no changed or removed lines. How is the previous pyshacl Shape method not executed now?

line 30 of store_service.py, I override the validate function of Shape
`Shape.validate = shape_validate`

lecriste:
> > I see only new lines, no changed or removed lines. How is the previous pyshacl Shape method not executed now?
> 
> line 30 of store_service.py, I override the validate function of Shape `Shape.validate = shape_validate`

And where is `Shape` used?

ssssarah:
Closed because another approach will be taken to fix this problem

PR #358, opened 2023-11-21
Use prepare uri when deprecating resources

ssssarah:
Extends [this MR](https://github.com/BlueBrain/nexus-forge/commit/02640967dd4034a4a4f5d566f322f2d9fdb9bd42), for the deprecation of schemas

### DIFF
--- a/kgforge/specializations/stores/bluebrain_nexus.py
+++ b/kgforge/specializations/stores/bluebrain_nexus.py
@@ -666,9 +666,16 @@ class BlueBrainNexus(Store):
         )
 
     def _deprecate_one(self, resource: Resource) -> None:
-        url = f"{self.service.url_resources}/_/{quote_plus(resource.id)}?rev={resource._store_metadata._rev}"
+
+        url, params = self.service._prepare_uri(resource)
+
         try:
             params_deprecate = copy.deepcopy(self.service.params.get("deprecate", None))
+
+            if params_deprecate is not None:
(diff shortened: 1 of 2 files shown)


kgforge/specializations/stores/bluebrain_nexus.py
```diff
@@ -666,9 +666,16 @@ def _deprecate_many(self, resources: List[Resource]) -> None:
         )
 
     def _deprecate_one(self, resource: Resource) -> None:
-        url = f"{self.service.url_resources}/_/{quote_plus(resource.id)}?rev={resource._store_metadata._rev}"
+
+        url, params = self.service._prepare_uri(resource)
+
         try:
             params_deprecate = copy.deepcopy(self.service.params.get("deprecate", None))
```
lecriste, line 673: Is this `self.service.params` the same of [this](https://github.com/BlueBrain/nexus-forge/pull/358/files#diff-9ee1d07daf4890e0d1087d9815baaea67a26d4fd6b31e546784b695165c8e68aR670) `params`?
If yes, I would replace `self.service.params` with `params`

ssssarah: No they're different, one comes from the initialisation of the service and the other is built in _prepare_uri

lecriste: OK, thanks.

lecriste (approved):

Issue #164, opened 2021-08-09
Non-specific error message on registering resources with expired token


eugeniashurko:
When the token expires and we are unauthorized, an error message is not specific enough, i.e. we get:

`JSONDecodeError: Expecting value: line 1 column 1`

ssssarah:
This has been adressed in https://github.com/BlueBrain/nexus-forge/pull/359

PR #359, opened 2023-11-22
Systematically raise for status on request and uniformize exception handling

ssssarah:
There were some calls to Delta that didn't `request.raise_for_status()` and so there would be an attempt to parse the body even though there wasn't one. 
This was noticed for example when a forge instance would have its token expire and some error codes in requests would not be checked, the body would be parsed, and a json error would be thrown. 

For all `requests.___` calls, the same `_catch_http_error` is being called. This also improves code readability to avoid having the the same try-except block repeated across the file. 

Adresses https://github.com/BlueBrain/nexus-forge/issues/164


lecriste (approved):
Looks good to me.

PR #362, opened 2023-11-28
Fix cyclic dependency between Mapping and KnowledgeGraphForge

ssssarah:
KnowledgeGraphForge picks specializations of mappings/mappers as default types of Mapping/Mapper to use. There's a cyclic dependency error if one tries to import both KnowledgeGraphForge and DictionaryMapping. 

This removes the imports of DictionaryMapping and DictionaryMapper in KnowledgeGraphForge: 
I think having core import specializations is not a good practice, and I've replaced the default value that KnowledgeGraphForge with the mapping and mapper defined by the store it holds. It could be that this makes no sense, if ever the intended usage of these mappings/mappers is different. 

I've also made the specializations file import the classes directly from the files they were defined, and not the packages they are contained in. I feel like it's more of a user interface feature to have the ability to import from packages. It could be argued that specializations should be defined by users and so they should also import the same way...


crisely09 (approved):
Seems to be fine.

PR #368, opened 2024-01-03
added timeouts to every requests. call

ssssarah:
There have been some issues recently in order to access some things hosted on github.io (ex: metadata context) (they should be fixed by the time of this review), and it's brought out one issue in forge: there were no timeouts when making requests, and some forge calls could hang for close to 10 minutes before getting an error that the server could not be reached.

This MR sets a timeout of 60 seconds for every request call in the codebase (constants are set in the different parts of forge that make requests, so that they may be changed depending on the use case)


crisely09:
Is there a reason not to have a global TIMEOUT variable, that could be overwritten for each case?
should it be accesible to be changed by the user?

ssssarah:
> Is there a reason not to have a global TIMEOUT variable, that could be overwritten for each case? should it be accesible to be changed by the user?

I would think this offers flexibility and that different implementations may go for their own timeout, fetching a file could have a different timeout than having a response with json in the body. 

The flexibility can still be kept by having each location have their own local variable that gets (for now) as its value the default one set somewhere else. Should one want to change its value, only the file/class local variable would need to be overridden, the request implementation wouldn't need to be changed and the global default variable wouldn't need to be changed either.

kgforge/core/commons/files.py
```diff
@@ -32,7 +36,7 @@ def load_file_as_byte(source: str):
         data = filepath.read_bytes()
     else:
         try:
-            response = requests.get(source)
+            response = requests.get(source, timeout=3600)
```
crisely09, line 39: I guess this is the only call that needs that much time?

ssssarah: It's a mistake, fixed it!

crisely09 (approved):

PR #370, opened 2024-01-09
Load local configuration when configuration tests

ssssarah:
When running tests that initialise the BlueBrainNexus store, load store configuration from `./examples/notebooks/use-cases/prod-forge-nexus.yml`

### DIFF
--- a/tests/specializations/stores/test_bluebrain_nexus.py
+++ b/tests/specializations/stores/test_bluebrain_nexus.py
@@ -22,6 +22,7 @@ import nexussdk
 import pytest
 from typing import Callable, Union, List
 
+from kgforge.core.commons.files import load_yaml_from_file
 from kgforge.core.resource import Resource
 from kgforge.core.archetypes.store import Store
 from kgforge.core.commons.context import Context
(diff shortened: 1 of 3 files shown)


crisely09:
Would it make sense to add here the change mentioned today: use a particular path as default to load a local configuration? instead of fetching the raw file from GitHub

ssssarah:
> Would it make sense to add here the change mentioned today: use a particular path as default to load a local configuration? instead of fetching the raw file from GitHub

This MR only deals with configurations used in tests. And it's only picking subparts of the configuration.
I think the default configuration thing should happen, but it doesn't feel relevant to this MR.

tests/specializations/stores/test_bluebrain_nexus.py
```diff
@@ -119,16 +120,38 @@ def registered_person(person, store_metadata_value):
...
-def nexus_store(context_project_patch, metadata_context_patch):
+def nexus_store(context_project_patch, metadata_context_patch, store_config):
+
+    store_config_cp = copy.deepcopy(store_config)
+
+    for key in ["endpoint", "model", "bucket", "file_resource_mapping", "token"]:
+        if key in store_config_cp:
+            store_config_cp.pop(key)
+
+    store_config.pop("endpoint")
```
MFSY, line 146: WHy pop endpoint here ?

ssssarah: My mistake, I should remove it as it's done in the first loop.

MFSY (approved):

PR #373, opened 2024-01-31
Pass view when forge.sparql, forge.elastic, forge.search call

ssssarah:
Feature to be able to specify a view when calling forge.sparql, forge.elastic and forge.search, instead of always using the one defined in the config

### DIFF
--- a/examples/notebooks/getting-started/04 - Querying.ipynb
+++ b/examples/notebooks/getting-started/04 - Querying.ipynb
@@ -17,7 +17,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": 141,
    "metadata": {
     "ExecuteTime": {
      "end_time": "2019-09-23T18:50:20.068658Z",
@@ -39,7 +39,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 2,
+   "execution_count": 142,
    "metadata": {},
    "outputs": [],
    "source": [
@@ -56,7 +56,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 3,
+   "execution_count": 143,
    "metadata": {},
    "outputs": [],
    "source": [
@@ -83,7 +83,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 4,
+   "execution_count": 144,
    "metadata": {},
    "outputs": [],
    "source": [
@@ -92,7 +92,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 5,
+   "execution_count": 145,
    "metadata": {},
    "outputs": [
     {
@@ -110,7 +110,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 6,
+   "execution_count": 146,
    "metadata": {},
    "outputs": [],
    "source": [
@@ -119,7 +119,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 7,
+   "execution_count": 147,
    "metadata": {},
    "outputs": [
     {
@@ -128,7 +128,7 @@
        "True"
       ]
      },
-     "execution_count": 7,
+     "execution_count": 147,
      "metadata": {},
      "output_type": "execute_result"
     }
@@ -147,7 +147,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 8,
+   "execution_count": 148,
    "metadata": {},
    "outputs": [],
    "source": [
@@ -156,7 +156,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 9,
+   "execution_count": 149,
    "metadata": {},
    "outputs": [
     {
@@ -174,7 +174,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 10,
+   "execution_count": 150,
    "metadata": {},
    "outputs": [
     {
@@ -192,7 +192,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 11,
+   "execution_count": 151,
    "metadata": {},
    "outputs": [],
    "source": [
@@ -201,7 +201,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 12,
+   "execution_count": 152,
    "metadata": {},
    "outputs": [
     {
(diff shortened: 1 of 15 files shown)


crisely09:
looks good, could you provide either a test or expand the query example notebook to use the `view` parameter?

ssssarah:
I would probably update the notebooks, as there are no unit tests for querying

kgforge/specializations/stores/bluebrain_nexus.py
```diff
@@ -303,7 +240,7 @@ def _upload_one(self, path: Path, content_type: str) -> Dict:
     # C[R]UD.
 
     def retrieve(
-            self, id_: str, version: Optional[Union[int, str]], cross_bucket: bool, **params
+            self, id_: str, version: Optional[Union[int, str]], cross_bucket: bool = False, **params
```
MFSY, line 243: This is done at forge.retrieve() level

kgforge/specializations/stores/bluebrain_nexus.py
```diff
@@ -704,19 +641,16 @@ def search(
...
         )
-        if search_endpoint not in [
-            self.service.sparql_endpoint["type"],
-            self.service.elastic_endpoint["type"],
-        ]:
-            raise ValueError(
-                f"The provided search_endpoint value '{search_endpoint}' is not supported. Supported "
-                f"search_endpoint values are: '{self.service.sparql_endpoint['type'], self.service.elastic_endpoint['type']}'"
-            )
-        if "filters" in params:
```
MFSY, line 717: Why remove this ?

kgforge/specializations/stores/bluebrain_nexus.py
ssssarah: The issue is that its superclass has the default value for cross_bucket set, and this one didn't so it led to warnings about the implementation having a different signature from its superclass. If I remove it here, I should remove it from the superclass

kgforge/specializations/stores/bluebrain_nexus.py
ssssarah: For the sake of usability, filters was *filters in the forge signature which made sense. However it was also *filters in the store signature, which doesn't make much sense in my opinion (users are not exposed to the store), so now it should be provided as a keyword parameter. So it's not possible for filters to ever make it to params, the provided value will be attributed to the explicit declaration of filters in the method signature

ssssarah: see https://github.com/BlueBrain/nexus-forge/commit/4d2580e4b7c4514278fa7385109cee0d794a6b36#diff-9ee1d07daf4890e0d1087d9815baaea67a26d4fd6b31e546784b695165c8e68aR686

examples/notebooks/getting-started/04 - Querying.ipynb
MFSY, line 1: In the section "Retrieving using the resorce url", cell 27 should use a resource id that will lead to the creation of a non existing resource in cell 28 and that will lead to the output of cell 31 to be true.

kgforge/specializations/stores/bluebrain_nexus.py
MFSY: Which warning and what is throwing that ?

kgforge/specializations/stores/bluebrain_nexus.py
MFSY: I see. This change from *filters to filters introduced an API change and it is not clear to me why this change was necessary. I'll review the why this change in another PR before release. For now, you are right to align the implementation.

MFSY:
> I would probably update the notebooks, as there are no unit tests for querying

Please provide unit tests for this change.

kgforge/specializations/stores/bluebrain_nexus.py
ssssarah: Come to think of it, this check can be kept but it would be done in forge method, not the store method

ssssarah:
> > I would probably update the notebooks, as there are no unit tests for querying
> 
> Please provide unit tests for this change.

There are no unit tests for querying currently, and this feature only checks the existence of an extra keyword argument, before doing the request.post call, so it's not even behavior that is isolated and can be tested through unit tests

kgforge/specializations/stores/bluebrain_nexus.py
ssssarah: The way I see *filters is nothing more than a way to simplify the usage of user-exposed functions. However this list unpacking was spread all the way down to parts of the code that are not user-exposed (ex: Sparql query builder) and it was preventing the passing of arguments in "any order" (with keywords). In a way, it was forcing each method call to be aware of the position of the arguments to be provided, and that was very error-prone (it's easier to provide parameters with keywords, that are often documented, and checked by an IDE, than it is to trust a developer to provide them in the right order, without keywords)

ssssarah:
I could make a unit test on https://github.com/BlueBrain/nexus-forge/blob/4fde40c8514769cb851bf620e3148586bae61920/kgforge/specializations/stores/nexus/service.py#L218 though

examples/notebooks/getting-started/04 - Querying.ipynb
ssssarah: I've pushed a fix, let me know if this is what you had in mind

kgforge/specializations/stores/bluebrain_nexus.py
MFSY: That could work. The API is what is in kgforge.core

examples/notebooks/getting-started/04 - Querying.ipynb
MFSY: Looks good.

kgforge/specializations/stores/bluebrain_nexus.py
ssssarah: The signature of the superclass defines the interface that any implementation should follow: all "promises" made at the superclass level should be kept by the implementations. If we want implementation-specific default values, then there shouldn't be a default value on the superclass: the subclass can have a default value without the superclass having one = specialised version of the superclass signature , but not the other way around = more general subclass than superclass

MFSY (approved):

PR #374, opened 2024-02-02
Update resource schema

ssssarah:
[DKE-1286](https://bbpteam.epfl.ch/project/issues/browse/DKE-1286)

### DIFF
--- a/kgforge/core/commons/execution.py
+++ b/kgforge/core/commons/execution.py
@@ -107,10 +107,19 @@ def catch_http_error(
         raise error_to_throw(error_message_formatter(e)) from e
 
 
-def run(fun_one: Callable, fun_many: Optional[Callable], data: Union[Resource, List[Resource]],
-        exception: Type[RunException], id_required: bool = False,
-        required_synchronized: Optional[bool] = None, execute_actions: bool = False,
-        monitored_status: Optional[str] = None, catch_exceptions: bool = True, **kwargs) -> None:
+def run(
+        fun_one: Callable,
+        fun_many: Optional[Callable],
+        data: Union[Resource, List[Resource]],
+        exception: Type[RunException],
+        id_required: bool = False,
+        required_synchronized: Optional[bool] = None,
+        execute_actions: bool = False,
+        monitored_status: Optional[str] = None,
+        catch_exceptions: bool = True,
+        **kwargs
+) -> None:
+
     # POLICY Should be called for operations on resources where recovering from errors is needed.
     if isinstance(data, List) and all(isinstance(x, Resource) for x in data):
         if fun_many is None:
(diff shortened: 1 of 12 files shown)


kgforge/specializations/stores/bluebrain_nexus.py
```diff
@@ -244,6 +253,30 @@ def _upload_one(self, path: Path, content_type: str) -> Dict:
 
     # C[R]UD.
 
+    @staticmethod
+    def _local_parse(id_value, version_params):
```
crisely09, line 257: maybe be completely explicit about what is parsed?  `_local_urlparse`?

kgforge/specializations/stores/bluebrain_nexus.py
```diff
@@ -522,7 +543,7 @@ def _update_one(self, resource: Resource, schema_id: str) -> None:
             metadata_context=None,
             context_resolver=self.service.resolve_context
         )
-        url, params = self.service._prepare_uri(resource, schema_id)
+        url, params = self.service._prepare_uri(resource, schema_id, keep_unconstrained=True)
```
crisely09, line 546: is `keep_unconstrained` really necessary? also, where can the user modify it?

ssssarah: _prepare_uri is used by most methods (register, tag, update...)
The issue is that in the case of update, and update schema, you'd like the unconstrained schema to be provided in the uri (to edit the schema for it to be unconstrained, or to keep it as is), but in the case of register, tag... you'd want an _, if ever the schema that is provided is unconstrained.json

kgforge/specializations/stores/bluebrain_nexus.py
ssssarah: Just renamed it :)

kgforge/specializations/stores/bluebrain_nexus.py
crisely09: I had one last though, maybe naming it `use_unconstrained_id`, to be more specific that you want to use the id?

ssssarah: Just changed it

kgforge/core/forge.py
```diff
@@ -737,6 +737,7 @@ def update(
 
         :param data: the resources to update
         :param schema_id: an identifier of the schema the updated resources should conform to
+        :param update_schema: whether the resource(s)' schemas should be updated to the schema_id value
```
lecriste, line 740: I don't see `update_schema` among the function arguments.

crisely09 (approved):

kgforge/core/commons/execution.py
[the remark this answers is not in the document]
ssssarah: I went over the code because it was being ran through the functions I used, and I improved the formatting because why not

kgforge/core/forge.py
ssssarah: My bad, initially I thought that both schema_id and a flag should be provided to enable schema change, but then it turns out only schema_id was necessary. I had made an implementation with the update_schema boolean, but then removed it and forgot to remove the docstring

PR #375, opened 2024-02-02
Fix filter exception on search

ssssarah:


kgforge/specializations/stores/bluebrain_nexus.py
```diff
@@ -627,11 +627,6 @@ def search(
             **params
     ) -> List[Resource]:
 
-        if "filters" in params:
```
crisely09, line 630: is this specific to ES views?

ssssarah: specifying filters through search used to be done with *filters, and in a previous MR, it was changed to filters. I changed it back to *filters but I forgot to change to the forge call to the store

crisely09: I see, yes, I remember the `*filters` argument.

MFSY (approved):

PR #377, opened 2024-02-07
resolving returns also the hasleafregionpart attribute of the resource

NicoRicardi:
forge.resolve now returns resource.hasLeafRegionPart if available. This helps when resolving brain regions to check if a layer should be provided or not.


crisely09 (approved):
Good with me

PR #379, opened 2024-02-08
Improved deflatten function

### LINKED ISSUE #378: the function deflatten in kgforge/core/conversions/dataframe.py assumes ordered data
`deflatten([('A.type', 'Atype'), ('A.something', 'Asomething'), ('B.type', 'Btype')], '.')` 
returns
`{'A': {'type': 'Atype', 'something': 'Asomething'}, 'B': {'type': 'Btype'}}`
while 
`deflatten([('A.type', 'Atype'), ('B.type', 'Btype'), ('A.something', 'Asomething')], '.')`
returns
{'A': {'something': 'Asomething'}}

NicoRicardi:
As of now, deflatten assumed that all items are ordered according to the nesting. Different input would lead to several values being discarded. For instance `[('p1.id', 'id1'), ('p1.type', 'type1'), ('p2.id', 'id2'), ('p2.type', 'type2'), ('p1.x', 'x1'), ('p2.x', 'x2')]` would cause such issue.
Furthermore, if a key appears both as nested and not nested (e.g. 'p1', and 'p1.x') the non-nested value would be discarded without providing any warning.

The current version ensures that:
- the output of flatten is independent of the order and no value is discarded
- a ValueError with an insightful message is raised if  a key appears both as nested and not nested

Two tests have been added to check this. 
Closes #378

### DIFF
--- a/kgforge/core/conversions/dataframe.py
+++ b/kgforge/core/conversions/dataframe.py
@@ -63,6 +63,7 @@ def _from_dataframe(row: Series, na: Union[Any, List[Any]], nesting: str) -> Res
     data = deflatten(items, nesting)
     return from_json(data, None)
 
+
 def deflatten(items: List[Tuple[str, Any]], sep: str) -> Dict:
     d = {}
     split = []
@@ -74,21 +75,16 @@ def deflatten(items: List[Tuple[str, Any]], sep: str) -> Dict:
             pre, _ = k1.split(sep, maxsplit=1)
             if pre in d:
                 raise ValueError(f'Mix of {pre} and {pre}{sep} (e.g. {k1}). Cannot be processed!')
-            l = []
+            pitems = []
             for n2, t2_ in enumerate(items):
-                try:
-                    k2, v2 = t2_
-                except:
-                    print(n2,t2_)
-                    return 0
+                k2, v2 = t2_
                 if isinstance(k2, str) and k2.startswith(f'{pre}{sep}'):
                     _, post = k2.split(sep, maxsplit=1)
-                    l.append((post, v2))
+                    pitems.append((post, v2))
                     split.append(n2)
-            d[pre] = deflatten(l, sep)
+            d[pre] = deflatten(pitems, sep)
         else:
             if k1 in d:
-                print('b',d)
                 raise ValueError(f'Mix of {pre} and {pre}{sep} (e.g. {k1}). Cannot be processed!')
             d[k1] = v1
     return d
--- a/kgforge/core/conversions/dataframe.py
+++ b/kgforge/core/conversions/dataframe.py
@@ -63,26 +63,32 @@ def _from_dataframe(row: Series, na: Union[Any, List[Any]], nesting: str) -> Res
     data = deflatten(items, nesting)
     return from_json(data, None)
 
-
 def deflatten(items: List[Tuple[str, Any]], sep: str) -> Dict:
     d = {}
-    i = 0
-    while i < len(items):
-        k, v = items[i]
-        if sep not in k:
-            d[k] = v
-            i += 1
+    split = []
+    for n1, t1_ in enumerate(items):
+        if n1 in split:
+            continue
+        k1, v1 = t1_
+        if isinstance(k1, str) and sep in k1:
+            pre, _ = k1.split(sep, maxsplit=1)
+            if pre in d:
+                raise ValueError(f'Mix of {pre} and {pre}{sep} (e.g. {k1}). Cannot be processed!')
+            l = []
+            for n2, t2_ in enumerate(items):
+                try:
+                    k2, v2 = t2_
+                except:
+                    print(n2,t2_)
+                    return 0
+                if isinstance(k2, str) and k2.startswith(f'{pre}{sep}'):
+                    _, post = k2.split(sep, maxsplit=1)
+                    l.append((post, v2))
+                    split.append(n2)
+            d[pre] = deflatten(l, sep)
         else:
-            pk, _, ck = k.partition(sep)
-            pitems = list(take_while(items[i:], f"{pk}{sep}"))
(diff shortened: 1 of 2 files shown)


tests/core/conversions/test_dataframe.py
```diff
@@ -64,7 +65,7 @@ def ndf():
 
 
 class TestConversionToDataFrame:
-
+
```
crisely09, line 68: trailing spaces are checked by the linting, the best is to also run them locally before pushing:
```
pycodestyle kgforge
pylint -j2 kgforge --fail-under=9.0
```

kgforge/core/conversions/dataframe.py
```diff
@@ -63,26 +63,32 @@ def _from_dataframe(row: Series, na: Union[Any, List[Any]], nesting: str) -> Res
...
+        if isinstance(k1, str) and sep in k1:
+            pre, _ = k1.split(sep, maxsplit=1)
+            if pre in d:
+                raise ValueError(f'Mix of {pre} and {pre}{sep} (e.g. {k1}). Cannot be processed!')
+            l = []
+            for n2, t2_ in enumerate(items):
+                try:
+                    k2, v2 = t2_
+                except:
+                    print(n2,t2_)
```
crisely09, line 82: you missed to remove your printings

kgforge/core/conversions/dataframe.py
```diff
@@ -63,26 +63,32 @@ def _from_dataframe(row: Series, na: Union[Any, List[Any]], nesting: str) -> Res
     data = deflatten(items, nesting)
     return from_json(data, None)
 
-
 def deflatten(items: List[Tuple[str, Any]], sep: str) -> Dict:
```
ssssarah, line 67: docstring?

crisely09 (dismissed):
Fine with me.

kgforge/core/conversions/dataframe.py
```diff
@@ -65,24 +65,51 @@ def _from_dataframe(row: Series, na: Union[Any, List[Any]], nesting: str) -> Res
...
+        Deflatten means that the separator implies nesting of the dictionary.
+
+    """
+    d = {}  # dictionary that will be returned
+    split = []  # those we have already split (we need to keep track)
+    for n1, t1_ in enumerate(items):
+        if n1 in split:  # already done, continue
+            continue
+        k1, v1 = t1_  # all tuples refer to a key-value pair in the DataFrame row
+        if isinstance(k1, str) and sep in k1:  # avoid issue if int or other class
```
ssssarah, line 97: I think this check that it is a string is also applicable to the else statement. So there should be an if that makes sure k1 is a str, otherwise an error is thrown. And then within that if you have your current if-else  based on the presence of the separator. 

Unless you would want something other than a string in your else statement?

ssssarah:
I'm not entirely sure why it is necessary to have nested iterations over items?

ssssarah:
https://stackoverflow.com/questions/54776916/inverse-of-pandas-json-normalize there could be some inspiration here?

ssssarah:
I've also not looked at it in enough detail but deflatten method seems to be used with rather "controlled" input -> extracted from a dataframe. 
So the cases for which there are ValueError could not be given as input. 
It really depends if we see someone use the function as it is and providing their own items, in which case, yes, you should check possibly faulty input. 
But here, it looks like it's an internal usage function with controlled input where these errors wouldn't be thrown.

kgforge/core/conversions/dataframe.py
NicoRicardi: well, no other function in the file has them, but still, I added it

kgforge/core/conversions/dataframe.py
NicoRicardi: Actually the is instance allows for non-strings to be processed in the else statement.
I do not think any resource should have non-string attributes, but I think that the error should not be raised by this function, but (if at all) by the Resource.from_json.
I see the following options:

- no check that it is a string => a non-string (e.g. int) would raise a generic error when we check if sep is in it. This is the behaviour of the master branch as of now
- check that this is a string in the if, not in the else => a non-string key in items is processed and returned in the dictionary. When the latter is passed to forge.from_json, it might raise an error
- check in both the if and the else => the non-string key and its value are discarded without any message. I do not think that this is desirable.
- we could raise a specific ValueError if any key in the key-value pairs is not a string

NicoRicardi:
> I've also not looked at it in enough detail but deflatten method seems to be used with rather "controlled" input -> extracted from a dataframe.
> So the cases for which there are ValueError could not be given as input.
> It really depends if we see someone use the function as it is and providing their own items, in which case, yes, you should check possibly faulty input.
> But here, it looks like it's an internal usage function with controlled input where these errors wouldn't be thrown.

There is no check - to the best of my understanding - that one does not have a column 'agent' and a column 'agent.something in their DataFrame'. As such the error makes sense

kgforge/core/conversions/dataframe.py
ssssarah: I would have an initial check that all keys are strings initially, without even getting into the details. 

`if not all(isinstance(k, str) for k, _ in items): raise something`

NicoRicardi:
> I'm not entirely sure why it is necessary to have nested iterations over items? If I understand correctly, you get the pre, and then when you have one pre, you iterate over everything trying to find everything that is a path starting in pre, but I would think there's a way to do it without such time complexity

At least at a first glance, the complexity is the same as the previous implementation. I feel like the problem seems easier at first, but I really do not see an easy way out of such complexity. I will look at the link you shared

ssssarah:
You have both nested loops and recursivity, the complexity is higher in this implementation

ssssarah:
> You have both nested loops and recursivity, the complexity is higher in this implementation

I missed a bit of code, indeed that was already the case

ssssarah (approved):

PR #383, opened 2024-02-15
Refactor batch request mechanism

ssssarah:
Motivated by:
Future batch retrieval feature, that could benefit from the existing mechanism, under some changes. + Aligning batch/single request logics (remove duplicate logic) 

For the methods: Deprecate, Tag, Update, Update Schema, Register:
- The logic of knowing which http method to use, which url to target, what query params to use, what headers to use, what payload to give and what exception to throw on error is prepared in `prepared_methods.py`. This information will either be used by the `requests.request` method (for synchronous calls, when performing these operations for one resource) or `aiohttp.ClientSession.request` (for asynchronous calls, when performing these operations for multiple resources) 

- When using `BatchRequestHandler.batch_request_for_resources`, the appropriate `prepare_x` should be provided. Everything else can either be found in the `service` instance or the `kwargs`. 

- `BatchRequestHandler.batch_request` provides a way to run a list of tasks. These tasks will be initialised using the `task_creator` parameter this method requires. 

- In the case of the methods cited above (deprecate, tag, update…), they all have the same `task_creator` method, which is defined by `BatchRequestHandler.create_tasks_for_resources`

In the effort to remove duplication of logic, some differences were spotted:
- Batch registration only targeted POST regardless of the presence of an id specified in the payload (https://bluebrainnexus.io//docs/delta/api/resources-api.html#create-using-post). This has been fixed, and now batch registration targets either POST or PUT, like single-resource registration did before (and still does). 
- The setting of the context after registration of a resource is done differently between single resource registration and multiple resource registration. https://github.com/BlueBrain/nexus-forge/blob/02da1021378ccd8bc231f45c8f40f941c6e282f5/kgforge/specializations/stores/bluebrain_nexus.py#L119 vs 
https://github.com/BlueBrain/nexus-forge/blob/02da1021378ccd8bc231f45c8f40f941c6e282f5/kgforge/specializations/stores/bluebrain_nexus.py#L190
This has not been addressed in this MR but should be in the future.


kgforge/specializations/stores/bluebrain_nexus.py
```diff
@@ -436,7 +412,12 @@ def retrieve(
             )
 
     def _retrieve_filename(self, id_: str) -> Tuple[str, str]:
-        response = requests.get(id_, headers=self.service.headers, timeout=REQUEST_TIMEOUT)
+        response = requests.request(
+            method=hdrs.METH_GET,
```
crisely09, line 416: why did you change the requests to the general method here? and keep `requests.get` in other places

kgforge/specializations/stores/bluebrain_nexus.py
```diff
@@ -124,80 +123,57 @@ def register_callback(task: Task):
...
-                params=params_register,
-                timeout=REQUEST_TIMEOUT
-            )
-        catch_http_error_nexus(response, RegistrationError)
+        data = payload
 
         response_json = response.json()
         resource.id = response_json["@id"]
         # If resource had no context, update it with the one provided by the store.
         if not hasattr(resource, "context"):
```
crisely09, line 174: about this id/context addition to the resource when registering... it doesn't seem like a extreme change to make one general callback, I would say ... go for it if you already thought of a way

kgforge/specializations/stores/nexus/prepare_methods.py
```diff
@@ -0,0 +1,188 @@
...
+
+    params_register = copy.deepcopy(service.params.get("register", {}))
+
+    identifier = resource.get_identifier()
+
+    url = Service.add_schema_and_id_to_endpoint(
+        service.url_resources, schema_id=schema_id, resource_id=identifier
+    )
+
+    context = service.model_context or service.context
```
crisely09, line 37: remind me, what did we say was the difference between the `model_context` and the `context`?

crisely09: the questions goes on the lines of ... if we are treating them as "the same", why do we have both? do we need them in different places?

kgforge/specializations/stores/bluebrain_nexus.py
ssssarah: Uh, I went for a find and replace streak, and I may have done it in places that weren't necessary, I can undo that one

kgforge/specializations/stores/nexus/prepare_methods.py
ssssarah: I don't know what the difference would be in the sense of use cases of having different contexts for the store and the model, but with respect to the configuration file (yaml), you have a context defined in your store, and you can define one (optionally) in the model. If no model is defined in the model, then the store model value is kept.

kgforge/specializations/stores/bluebrain_nexus.py
ssssarah: What do you mean by a general callback?

kgforge/specializations/stores/bluebrain_nexus.py
crisely09: hmm maybe make them all consistent? I don't know ...

kgforge/specializations/stores/bluebrain_nexus.py
crisely09: there is a default callback method, that is shared within all the possible resource tasks, that is what I meant, why not to move the option of adding id/context there, to have one single method?

kgforge/specializations/stores/nexus/prepare_methods.py
crisely09: hmm OK

kgforge/specializations/stores/bluebrain_nexus.py
ssssarah: The setting of the resource.id and context look harmless at first glance so I could see them happening for any resource operation. But are they really harmless?

First there's the question of harmonising the register_one and register_many callbacks before even thinking of making them the same as the other operations'

kgforge/specializations/stores/bluebrain_nexus.py
ssssarah: They were definitely motivated by something for some methods, and not for others. I could keep the ones with valid motivation as request.request and revert back to the specialised method for the others

ssssarah: I've undone the ones that weren't necessary

crisely09:
@MFSY ?

MFSY (approved):

PR #387, opened 2024-02-21
Set jsonld context version to 1.1

### LINKED ISSUE #386: Context does not resolve NCBITaxon terms.
Resolving an `NCBITaxon` term using forge's context:
```python
c = forge.get_model_context()
c.resolve("NCBITaxon:10000")
```
Returns:
```
'ncbitaxon:10000'
```
Failing to resolve the prefix.

In contrast when an mba term is resolved it works as expected:
```python
resolved = c.resolve("mba:997")
assert resolved == 'http://api.brain-map.org/api/v2/data/Structure/997'
```
The `NCBITaxon` term is present in the context's document:
```python
c.document["@context"]["NCBITaxon"]
# 'http://purl.obolibrary.org/obo/NCBITaxon_'
```
Checking how the `mba` and `NCBITaxon` terms are stored in the context:
```
Term(id='http://purl.obolibrary.org/obo/NCBITaxon_', name='NCBITaxon', type=0, container=set(), index=None, language=0, reverse=False, context=0, prefix=False, protected=False)
Term(id='http://api.brain-map.org/api/v2/data/Structure/', name='mba', type=0, container=set(), index=None, language=0, reverse=False, context=0, prefix=True, protected=False)
```
There is a difference in the `prefix` boolean, which is False in the `NCBITaxon` case. That leads to not replacing it in this line:
https://github.com/RDFLib/rdflib/blob/6563ea0a4ced5b3c858fe143bf3360cc1c1eb711/rdflib/plugins/shared/jsonld/context.py#L356

Most of the times this is not an issue for forge because it merges the source of a resource with the metadata and iin most cases the id in the source is expanded. However, in a resource where the term is also collapsed in the source, then the resolution fails:
```python
r = forge.retrieve("https://bbp.epfl.ch/data/bbp/mmb-point-neuron-framework-model/9a107d17-3b9c-4bf6-a68a-c822ffa79c75")
r.subject.species
# Resource(_last_action=None, _validated=False, _synchronized=True, _store_metadata=None, id='ncbitaxon:10090', label='Mus musculus', _inner_sync=True)
```

MFSY:
* this enables non IRI-delimiting character not present in rdflib.plugins.shared.jsonld.context.URI_GEN_DELIMS (e.g '_' in `"NCBITaxon": "http://purl.obolibrary.org/obo/NCBITaxon_"`) to be used when defining jsonld context prefix

As mentioned in https://github.com/RDFLib/rdflib/issues/2606, `@prefix` could be used to signal prefixes to jsonld context 1.1. E.g:
`"NCBITaxon": {"@id": "http://purl.obolibrary.org/obo/NCBITaxon_", "@prefix": True}` 

Fixes #386


setup.py
```diff
@@ -35,33 +35,25 @@
...
+    python_requires=">=3.8.1",
+    setup_requires=["setuptools_scm"],
     install_requires=[
         "hjson",
         "pyyaml",
         "pandas",
         "nexus-sdk",
         "aiohttp",
-        "rdflib==6.3.2",
+        "rdflib==7.0.0",
```
eleftherioszisis, line 46: It seems that in spack we have up until 6.3.2:
https://github.com/BlueBrain/spack/blob/70cb1d988a88052205672d05984f87470afafd9c/var/spack/repos/builtin/packages/py-rdflib/package.py#L24

MFSY: Good catch. Not sure how easy is it to add 7.0.0 there. Is changing the following enough:

- `version("7.0.0", sha256="9995eb8569428059b8c1affd26b25eac510d64f5043d9ce8c84e0d0036e995ae")` from https://pypi.org/project/rdflib/#files
Also for pyshacl:
- version("0.25.0",sha256="91e87ed04ccb29aa47abfcf8a3e172d35a8831fce23a011cfbf35534ce4c940b")` in [py-pyshacl/package.py](https://github.com/BlueBrain/spack/blob/70cb1d988a88052205672d05984f87470afafd9c/var/spack/repos/builtin/packages/py-pyshacl/package.py) from https://pypi.org/project/pyshacl/#modal-close

And also changing the python requirement there.

If not that simple, then we can do the version bumps at next forge release as it is not required for having jsonld 1.1

eleftherioszisis: Given that these packages are builtins and not from BlueBrain a patch will need to be added. See:
https://github.com/BlueBrain/spack/blob/develop/bluebrain/documentation/packages.md#adding-new-versions-or-modifying-packages-with-non-bluebrain-code

crisely09 (approved):
Looks good to me.

Issue #388, opened 2024-02-26
Missing forge method to add images to resources


crisely09:
Similar to `add_distribution` to the dataset objects, we should have a `add_image` for Resources.
This would be very useful, for example for traces, but now E-Models will also have images (to be used for thumbnails), and maybe other model entities would require the same.

MFSY:
Can this be closed @crisely09 ?

crisely09:
Yes, this can be closed.

PR #389, opened 2024-02-26
Add `add_image` method to `Dataset` entity

crisely09:
To tackle [#388](https://github.com/BlueBrain/nexus-forge/issues/388)

This is the simplest implementation to "attach" and image, and then use it to add it to any resource, similar to `attach` for distributions.

Example of how it is used can be found inside the 03 - Storing notebook.
The way to add the image would be :
```
image = forge.attach_image(path, content_type, about)
resource.image = image
```


eleftherioszisis:
How would it work in the case of multiple images? Just by stacking multiple lazy actions together like:
```python
image = forge.attach_image(path, content_type, about)
resource.image = [forge.attach_image(path, content_type, about) for path in paths]
```
?

MFSY:
I think this capability should go to Resource or one of its descendants just like Dataset currently holds specific data models for distributions and provenance.

crisely09:
> I think this capability should go to Resource or one of its descendants just like Dataset currently holds specific data models for distributions and provenance.

@MFSY  my issue with adding it to `Resource` is that we would need to pass `forge` to it, as it is done in `Dataset`, and that is kind of ugly. Also because that would mean a big change in how we have been initializing `Resource`. 
Maybe passing forge only to the method `add_image` ?

MFSY:
> > I think this capability should go to Resource or one of its descendants just like Dataset currently holds specific data models for distributions and provenance.
> 
> @MFSY my issue with adding it to `Resource` is that we would need to pass `forge` to it, as it is done in `Dataset`, and that is kind of ugly. Also because that would mean a big change in how we have been initializing `Resource`. Maybe passing forge only to the method `add_image` ?
I am okay with having the method accept a forge object. Now Resource can also be extended with an Entity class that will have this method which can accept forge. In that case, client will need to create an Entity from a resource before using the method.

AurelienJaquier:
> > How would it work in the case of multiple images? Just by stacking multiple lazy actions together like:
> > ```python
> > resource.image = [forge.attach_image(path, content_type, about) for path in paths]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ?
> 
> In fact, the path can be a directory, and that would take all the files inside, and make the list automatically

What if we have files from different directories? Would the piece of code provided by elefterios work?

crisely09:
> > > How would it work in the case of multiple images? Just by stacking multiple lazy actions together like:
> > > ```python
> > > resource.image = [forge.attach_image(path, content_type, about) for path in paths]
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > ?
> > 
> > 
> > In fact, the path can be a directory, and that would take all the files inside, and make the list automatically
> 
> What if we have files from different directories? Would the piece of code provided by elefterios work?

I think it should, yes

darshanmandge:
Hi @crisely09!
Can this PR be merged now?

stefanoantonel:
Any update on this merge?

MFSY:
Hi guys,
We'll look into this this week.

crisely09:
With the current changes, to add an image, one has to start from a dataset:
```
dataset = Dataset.from_resource(forge, resource)
dataset.add_image(path, content_type, about)
```
This should work in the same way as explained before for paths and for a list of files.

PR #393, opened 2024-03-12
Fixes deep nodeshape path eval error

MFSY:
Currently the pySHACL [throws](https://github.com/RDFLib/pySHACL/blob/master/pyshacl/shape.py#L468) a `ReportableRuntimeError("Evaluation path too deep!\n{}".format(path_str))`  exception when evaluating a shape if the length of its transitive closure of the sh:node property is bigger or equal to 30.  

Give a node shape, this PR addresses this by:
* first recursively collecting all the property shapes directly defined (through sh:property) by the node shape or inherited from the node shape parents 
* then link those collected property shapes to the node shape through sh:property
* finally remove the node shape <-> parent shape relatioinships

Along the way, the expected data model for a shacl shape is aligned between the RDF StoreService and DirectoryService:

```json
{
 "@context":"https://acontext"
  "@id": "http://shapes.ex/employee",
  "@type": "nxv:Schema",
  "owl:imports": [
    "http://shapes.ex/person"
  ],
  "shapes": [
    {
     "@id": "this:EmployeeShape",
      "@type": "sh:NodeShape",
       ....
    }
  ]
}
```


crisely09:
The problem is due to pytest incompatibility, Version: 7.2.1 works fine. I think we should move away from pytest-bdd, and re-format the tests, but not in this PR.

MFSY:
> The problem is due to pytest incompatibility, Version: 7.2.1 works fine. I think we should move away from pytest-bdd, and re-format the tests, but not in this PR.

What do you recommend ? get pytest 7.2.1 in setup ? Is t what work for you ?

crisely09:
Yes, locally I used that version and it works. At least for now we could setup forge with pytest==7.2.1,  while we make the changes in the tests to remove `pytest-bdd`

kgforge/specializations/models/rdf/service.py
```diff
@@ -128,16 +141,22 @@ def lookup_shape_from_node(self, node: URIRef) -> Shape:
 
 class RdfService:
 
-    def __init__(self, graph: Graph, context_iri: Optional[str] = None) -> None:
+    def __init__(self, graph: GraphLike, context_iri: Optional[str] = None) -> None:
 
         if context_iri is None:
             raise ConfigurationError("RdfModel requires a context")
         self._graph = graph
+        self.NXV = Namespace("https://bluebrain.github.io/nexus/vocabulary/")
```
crisely09, line 149: this is very Nexus specific, I am not convinced it should be part of the RdfService

MFSY: It is a namespace defined and used by Nexus Delta for its vocab (NXV.deprecated, NXV.createdBy, NXV.shapes, ...). This PR aligned the format for shacl expected by all rdf services. There is now a need for an id of type NXV.Schema which can point to one or many shacl shapes through NXV.shapes. So while NXV namespace was defined by Nexus Delta in the first place, it is here just a namespace needed to define schemas. It's usage does not implies using BlueBrainNexusStore.

crisely09: But any namespace used ... it should not be something hard-coded, it should be something configurable, no?

kgforge/specializations/models/rdf/store_service.py
```diff
@@ -68,93 +74,96 @@ def resolve_context(self, iri: str) -> Dict:
...
+                query, debug=False, limit=limit, offset=offset
+            )
             for r in resources:
                 shape_uri = URIRef(r.shape)
-                class_to_shapes[r.type] = shape_uri
-                shape_resource[shape_uri] = URIRef(r.resource_id)
+                class_to_shape[r.type] = shape_uri
+                shape_to_defining_resource[shape_uri] = URIRef(r.resource_id)
+                shape_to_named_graph[shape_uri] = URIRef(r.resource_id + "/graph")
+                defining_resource_to_named_graph[URIRef(row["resource_id"])] = URIRef(
```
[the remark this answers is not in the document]
crisely09: it should be `r.resource_id` instead ?

kgforge/specializations/models/rdf/service.py
MFSY: Not always I might say. There is a balance to find. NXV.shapes or OWL.imports in [here](https://github.com/BlueBrain/nexus-forge/pull/393/files#diff-20fe56168eb41def80c78aa3d4c1f0dbeddb920f8859a795c86f3f148cc4c408R302) can't be changed (for the servie to work) removing the necessity to configure that for now. If in the future we want to support use defined schema format in addition to shacl, the configuration would definitely make sense

kgforge/specializations/models/rdf/directory_service.py
```diff
@@ -61,30 +66,47 @@ def resolve_context(self, iri: str) -> Dict:
...
+            context=self.context,
+        )
         res = self._graph.query(query)
-        return {row["type"]: row["shape"] for row in res}
+        class_to_shape = {}
+        shape_to_defining_resource = {}
+        shape_to_named_graph = {}
+        defining_resource_to_named_graph = {}
+        for row in res:
+            class_to_shape[URIRef(row["type"])] = URIRef(row["shape"])
```
crisely09, line 83: small question: why to use an `URIRef` object instead of the `uri` as it is?

crisely09: also, the types can be loaded in shrink or expanded form, I think this should be uniformized here, using the `self.context.expand()` method

kgforge/specializations/models/rdf/directory_service.py
```diff
@@ -61,30 +66,47 @@ def resolve_context(self, iri: str) -> Dict:
...
+            )
+        return (
+            class_to_shape,
+            shape_to_defining_resource,
+            shape_to_named_graph,
+            defining_resource_to_named_graph,
+        )
+
+    def load_shape_graph(self, graph_id, schema_id):
+        return self._graph.graph(rdflib.term.URIRef(graph_id))
```
crisely09, line 99: change `rdflib.term.URIRef` to `URIRef` as it is already imported that way, to be consistent with previous lines

kgforge/specializations/models/rdf/store_service.py
```diff
@@ -28,37 +31,40 @@
...
+        self.store_metadata_iri = (
+            self.default_store.service.store_context
+            if hasattr(self.default_store.service, "store_context")
             else Namespace(Service.NEXUS_CONTEXT_FALLBACK)
-        self._shapes_to_resources: Dict
-        self._imported = []
-        self._graph = Graph()
+        )
+        self._graph = rdflib.Dataset()
         self._sg = ShapesGraphWrapper(self._graph)
```
crisely09, line 50: as self._sg is used within RdfService, probably it should be initialized there

kgforge/specializations/models/rdf/service.py
```diff
@@ -162,18 +181,20 @@ def validate(self, resource: Resource, type_: str):
...
-                shape_iri = self.types_to_shapes[type_]
+            type_to_validate = type_ if type_ else resource.type
+            shape_iri = self.get_shape_iri_from_class_fragment(type_to_validate)
         except AttributeError as exc:
-            raise TypeError("Resource requires a type attribute") from exc
-
+            raise TypeError(
+                "Resource requires a type attribute or a type_ parameter should be provided"
+            ) from exc
         data_graph = as_graph(resource, False, self.context, None, None)
```
crisely09, line 190: this line assumes that the resource context is the same as the given context, while this is not the case when loading the model from the directory and validating a resource with a context from a store

kgforge/specializations/models/rdf/directory_service.py
MFSY: Done

kgforge/specializations/models/rdf/store_service.py
MFSY: The RdfService delegates the building of _sg to specific type of services.

kgforge/specializations/models/rdf/directory_service.py
MFSY: >also, the types can be loaded in shrink or expanded form, I think this should be uniformized here, using the self.context.expand() method

The directory service gets a sparql result so all uris are already expanded. The directory service on the other hand needs that.

>small question: why to use an URIRef object instead of the uri as it is?

To catch as early as possible malformed URI

kgforge/specializations/models/rdf/store_service.py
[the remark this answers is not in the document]
MFSY: Done

kgforge/specializations/models/rdf/directory_service.py
```diff
@@ -15,25 +15,28 @@
...
-        self._graph = load_rdf_files(dirpath)
+
+        self._graph = _load_rdf_files_as_graph(dirpath)
         self._sg = ShapesGraphWrapper(self._graph)
         super().__init__(self._graph, context_iri)
 
     def schema_source_id(self, schema_iri: str) -> str:
-        # FIXME should return the file path where the schema is in
-        return schema_iri
+        return self.shape_to_defining_resource[schema_iri]
```
ssssarah, line 39: This method is supposed to say where the schema has been defined so it should refer to the path of the file that was loaded that contains the schema

kgforge/specializations/models/rdf/service.py
MFSY: Yes. The service is not making clear that self.context should be the one used for resources to validate. I recommend to have that in a follow up PR

kgforge/core/commons/sparql_query_builder.py
```diff
@@ -345,20 +367,111 @@ def _replace_in_sparql(
...
 
+    @staticmethod
+    def create_select_query(
+        vars_: List[str],
+        statements: Union[str, List[str]],
+        distinct: bool,
+        search_in_graph: bool,
+        union: bool = False,
+    ):
+        if not union:
```
ssssarah, line 400: Maybe you should check that statements is a string if not union? 
So a nested if and if it's not the case raise an Error that statements should be a string.

I would see it as an if that check union the boolean, and after that just an else case, and then within those, if statements.

if not union: (if not isinstance(statements, str) raise Ex, else ...)
else: (if not isinstance(statements, list) raise Ex, else ...)

kgforge/core/commons/sparql_query_builder.py
```diff
@@ -345,20 +367,111 @@ def _replace_in_sparql(
...
+        else ""
+    )
+    if optional and deprecated_property_uri:
+        deprecated_statement = " OPTIONAL {" + deprecated_statement + "}"
+    if optional and defining_property_uri:
+        defining_statement = " OPTIONAL {" + defining_statement + "}"
+
+    shape_target_statements = "?shape sh:targetClass ?type"
+    shape_node_statements = "SELECT (?shape as ?type) ?shape ?resource_id WHERE { ?shape a sh:NodeShape . ?shape a rdfs:Class"
+    extra_statements = [defining_statement] + [deprecated_statement]
```
ssssarah, line 451: why not declare a single array with two values

kgforge/core/commons/sparql_query_builder.py
```diff
@@ -345,20 +367,111 @@ def _replace_in_sparql(
...
+        f"?resource_id <{defining_property_uri}> ?shape"
+        if defining_property_uri
+        else ""
+    )
+    if optional and deprecated_property_uri:
+        deprecated_statement = " OPTIONAL {" + deprecated_statement + "}"
+    if optional and defining_property_uri:
+        defining_statement = " OPTIONAL {" + defining_statement + "}"
+
+    shape_target_statements = "?shape sh:targetClass ?type"
```
ssssarah, line 449: singular, there's just one statement?

kgforge/core/commons/sparql_query_builder.py
```diff
@@ -345,20 +367,111 @@ def _replace_in_sparql(
...
+    shape_node_statements = "SELECT (?shape as ?type) ?shape ?resource_id WHERE { ?shape a sh:NodeShape . ?shape a rdfs:Class"
```
ssssarah, line 450: looks like more than a statement? Maybe shape_node_select_subquery?

kgforge/core/commons/sparql_query_builder.py
```diff
@@ -345,20 +367,111 @@ def _replace_in_sparql(
...
+    extra_statements = [defining_statement] + [deprecated_statement]
+    if statements:
+        extra_statements.extend(statements)
+    shape_target_statements_str = ".".join([shape_target_statements] + extra_statements)
+    shape_node_statements_str = (
```
ssssarah, line 455: For the sake of readability, would it be possible to make shape_node_statements a template string in which you add 
"." + ".".join(extra_statements)

kgforge/core/commons/sparql_query_builder.py
```diff
@@ -345,20 +367,111 @@ def _replace_in_sparql(
...
 
 def _box_value_as_full_iri(value):
     return f"<{value}>" if is_valid_url(value) else value
+
+
+def build_shacl_query(
+    statements: List[str] = None,
+    defining_property_uri: str = None,
+    deprecated_property_uri: str = None,
+    optional: bool = False,
```
ssssarah, line 430: Why is optional controlling both defining_property_uri and deprecated_property_uri? 

Why would someone want to include both of these, or not include them? Why is their optional status tied (both or none are optional)

kgforge/core/commons/sparql_query_builder.py
```diff
@@ -345,20 +367,111 @@ def _replace_in_sparql(
...
+
 
 def _box_value_as_full_iri(value):
     return f"<{value}>" if is_valid_url(value) else value
+
+
+def build_shacl_query(
+    statements: List[str] = None,
+    defining_property_uri: str = None,
+    deprecated_property_uri: str = None,
```
ssssarah, line 429: What's the point of including the deprecated statement if you don't add it to vars_?

kgforge/specializations/models/rdf/service.py
ssssarah: Can't this be propagated from the Store/vocabulary/namespace config entry?

ssssarah: I don't understand why the code you've pointed to is not compatible with a namespace provided on service initialisation

kgforge/specializations/models/rdf/service.py
```diff
@@ -187,23 +209,18 @@ def generate_context(self) -> Dict:
         raise NotImplementedError()
 
     @abstractmethod
-    def _build_shapes_map(self) -> Dict:
-        """Queries the source and returns a map of owl:Class to sh:NodeShape"""
+    def _build_shapes_map(self) -> Tuple[Dict, Dict, Dict]:
```
ssssarah, line 212: Looks like a Dict is missing. Both implementations return 4 dictionaries. The return type should change + the doc

kgforge/specializations/models/rdf/directory_service.py
ssssarah: I've gone into the details of this in this MR: https://github.com/BlueBrain/nexus-forge/pull/351/files#diff-defd061221b05700c81209720e0efdc6ce1fdb14edc68e93a926d8f1e7af8a6eR59  

But the problem is that in the former "load_rdf_files" (now named _load_rdf_files_as_graph), the record of which file led to which graph should be kept, and used here

kgforge/specializations/models/rdf/service.py
```diff
@@ -162,18 +181,20 @@ def validate(self, resource: Resource, type_: str):
...
+            ) from exc
         data_graph = as_graph(resource, False, self.context, None, None)
-        return self._validate(shape_iri, data_graph)
+        shape, shacl_graph = self.get_shape_graph(shape_iri)
+        return self._validate(shape_iri, data_graph, shape, shacl_graph)
 
     @abstractmethod
-    def _validate(self, iri: str, data_graph: Graph) -> Tuple[bool, Graph, str]:
+    def _validate(
+        self, iri: str, data_graph: Graph, shape: Shape, shacl_graph: Graph
```
ssssarah, line 202: Both implementations (StoreSrrvice and DirectoryService) do not have shacl_graph as an extra parameter.

kgforge/specializations/models/rdf/service.py
```diff
@@ -162,18 +181,20 @@ def validate(self, resource: Resource, type_: str):
                     "Resource has list of types as attribute and type_ parameter is not specified. "
                     "Please provide a type_ parameter to validate against it."
                 )
-            if type_ is None:
-                shape_iri = self.types_to_shapes[resource.type]
-            else:
-                shape_iri = self.types_to_shapes[type_]
+            type_to_validate = type_ if type_ else resource.type
```
ssssarah, line 190: 1. I would expect that even if a resource does not have a type attribute, one could validate if a type_ was provided. This is not the case because we check whether resource.type is a list, and that would always raise an AttributeError, even if type_ was not None. So first type_ is None should be checked, then in a nested if, resource.type being a list should be checked. 

2. shape_iri = self.get_ ... should not be within the try block

kgforge/specializations/models/rdf/service.py
```diff
@@ -272,3 +291,108 @@ def traverse_properties(properties) -> Tuple[Dict, Dict]:
...
+            self._sg = ShapesGraphWrapper(self._graph)
+            shape = self._sg.lookup_shape_from_node(shape_uriref)
+        return shape, shape_graph
+
+    def get_shape_iri_from_class_fragment(self, fragment):
+        try:
+            type_expanded_cls = self.context.expand(fragment)
+            return self.class_to_shape[URIRef(type_expanded_cls)]
+        except KeyError as ke:
+            raise TypeError(f"Unkown type: {ke}") from ke
```
ssssarah, line 402: Typo

kgforge/specializations/models/rdf/service.py
```diff
@@ -187,24 +208,15 @@ def generate_context(self) -> Dict:
         raise NotImplementedError()
 
     @abstractmethod
-    def _build_shapes_map(self) -> Dict:
-        """Queries the source and returns a map of owl:Class to sh:NodeShape"""
+    def _build_shapes_map(self) -> Tuple[Dict, Dict, Dict]:
+        """
+        Queries the source and returns as tuple:
+        * a Dict of a targeted owl:Class to a shape and
+        * a Dict of a shape to the resource definining it
```
ssssarah, line 221: typo

kgforge/specializations/models/rdf/service.py
```diff
@@ -272,3 +291,108 @@ def traverse_properties(properties) -> Tuple[Dict, Dict]:
...
+        sh_nodes.extend(list(self._graph.objects(node_shape_uriref, SH.node)))
+        for s in schema_defines_shape:
+            sh_nodes.extend(list(self._graph.objects(node_shape_uriref, s / SH.node)))
+            sh_properties.extend(
+                list(self._graph.objects(node_shape_uriref, s / SH.property))
+            )
+        for sh_node in set(sh_nodes):
+            if (
+                str(sh_node) != str(node_shape_uriref)
+                and str(sh_node) != "http://www.w3.org/1999/02/22-rdf-syntax-ns#nil"
```
ssssarah, line 355: Why not use the RDF namespace?

kgforge/specializations/models/rdf/store_service.py
```diff
@@ -28,37 +31,43 @@
...
-        self.store_metadata_iri = self.default_store.service.store_context if hasattr(self.default_store.service, "store_context") \
+        self.store_metadata_iri = (
+            self.default_store.service.store_context
+            if hasattr(self.default_store.service, "store_context")
             else Namespace(Service.NEXUS_CONTEXT_FALLBACK)
-        self._shapes_to_resources: Dict
-        self._imported = []
-        self._graph = Graph()
+        )
+        self._graph = rdflib.Dataset()
```
ssssarah, line 49: Similarly to my comment on the directory service, the rdflib.Dataset() instance should be passed to the super's constructor so that it sets self._graph.
With respect to self._sg needing the dataset instance, I would suggest that self._sg be initialized in the superclass, also because in RdfService code, there is an assumption made that self._sg exists, even though it isn't initialized in its constructor (all its subclasses initialise it but this isn't safe)

kgforge/specializations/models/rdf/service.py
```diff
@@ -128,21 +140,34 @@ def lookup_shape_from_node(self, node: URIRef) -> Shape:
...
+            self.class_to_shape,
+            self.shape_to_defining_resource,
+            self.shape_to_named_graph,
+            self.defining_resource_to_named_graph,
+        ) = self._build_shapes_map()
+        self._imported = []
 
-    def schema_source_id(self, schema_iri: str) -> str:
+    @abstractmethod
+    def schema_source_id(self, shape_uri: str) -> str:
```
ssssarah, line 163: I think this shouldn't be abstract and this should have 
return self.shape_to_defining_resource[schema_iri]
Which is what both implementations currently have.
I don't see why any additional logic could be envisioned that would justify having the subclasses implement it

kgforge/specializations/models/rdf/directory_service.py
crisely09: I just confirmed, the directory service could also get the shrink versions of the uris, the expansion needs to happen there too

kgforge/specializations/models/rdf/service.py
```diff
@@ -272,3 +301,102 @@ def traverse_properties(properties) -> Tuple[Dict, Dict]:
...
+
+    def get_shape_graph(self, shape_uriref: URIRef) -> Tuple[Shape, Graph]:
+        try:
+            shape = self._sg.lookup_shape_from_node(shape_uriref)
+            shape_graph = self.transitive_load_shape_graph(
+                self.shape_to_named_graph[shape_uriref],
+                self.shape_to_defining_resource[shape_uriref],
+                shape_uriref,
+            )
+        except KeyError:
```
ssssarah, line 386: What's the reason for trying again the transitive_load_shape_graph? 
Is it transitive_load_shape_graph that can raise the KeyError or is it lookup_shape_from_node?

kgforge/specializations/models/rdf/service.py
```diff
@@ -272,3 +301,102 @@ def traverse_properties(properties) -> Tuple[Dict, Dict]:
         context.update({key: terms[key] for key in sorted(terms)})
 
         return {"@context": context} if len(context) > 0 else None
+
+    def transitive_load_shape_graph(
```
ssssarah, line 305: I will review this one later 😰

kgforge/specializations/models/rdf/store_service.py
MFSY: self._sg init is now moved to RDFService

kgforge/core/commons/sparql_query_builder.py
MFSY: done

kgforge/core/commons/sparql_query_builder.py
MFSY: done

kgforge/core/commons/sparql_query_builder.py
MFSY: Deprecated should be false if deprecated_property_uri  is provided. No need for variable

kgforge/core/commons/sparql_query_builder.py
MFSY: done

kgforge/specializations/models/rdf/directory_service.py
MFSY: Both done

kgforge/core/commons/sparql_query_builder.py
MFSY: ?resource_id is no longer optional.

kgforge/specializations/models/rdf/service.py
MFSY: done

kgforge/specializations/models/rdf/store_service.py
```diff
@@ -28,37 +31,43 @@
...
         self._sg = ShapesGraphWrapper(self._graph)
         super().__init__(self._graph, context_iri)
 
-    def schema_source_id(self, schema_iri: str) -> str:
-        return self._shapes_to_resources[schema_iri]
+    def schema_source_id(self, shape_uri: str) -> str:
+        return str(self.shape_to_defining_resource[URIRef(shape_uri)])
+
+    def schema_id(self, shape_uri: str) -> str:
+        return self.schema_source_id(shape_uri)
```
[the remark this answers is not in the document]
MFSY: Schema_id is the defining resource in both store and directory for now but the schema source id is different: 
- directory: file path
- store: schema_id+"/graph" to avoid named graph and schema to have the same id in the RDFDataset
So both services can't share the same implementation of schema_source_id. For schema_id, they can share it for now.

kgforge/specializations/models/rdf/service.py
MFSY: schema_source_id has different implementation in the services. schema_id() is the one with the same implementation currrently

kgforge/core/commons/sparql_query_builder.py
MFSY: Done

kgforge/specializations/models/rdf/directory_service.py
MFSY: Okay. Still surprised that a SPARQL query results returned curies. Can you print past here the SPARQL result json output you are getting ?

crisely09:
```
(rdflib.term.URIRef('bmo:ExtractionTargetsConfiguration'), rdflib.term.URIRef('https://bbp.epfl.ch/shapes/dash/extractiontargetsconfiguration/shapes/ExtractionTargetsConfigurationShape'), rdflib.term.URIRef('https://bbp.epfl.ch/shapes/dash/extractiontargetsconfiguration'), rdflib.term.URIRef('shapes/datashapes/core/extractiontargetsconfiguration.json'))
(rdflib.term.URIRef('https://neuroshapes.org/BrainSlicing'), rdflib.term.URIRef('https://neuroshapes.org/dash/brainslicing/shapes/BrainSlicingShape'), rdflib.term.URIRef('https://neuroshapes.org/dash/brainslicing'), rdflib.term.URIRef('shapes/test_tmp_dir/ontologies/bbp/shapes_jsonld_expanded/datashapes/core/brainslicing.ttl'))
(rdflib.term.URIRef('https://neuroshapes.org/IonChannelGene'), rdflib.term.URIRef('https://neuroshapes.org/dash/ionchannelgene/shapes/IonChannelGeneShape'), rdflib.term.URIRef('https://neuroshapes.org/dash/ionchannelgene'), rdflib.term.URIRef('shapes/test_tmp_dir/ontologies/shapes_jsonld_expanded/datashapes/core/ionchannelgene.ttl'))
(rdflib.term.URIRef('https://neuroshapes.org/DetailedCircuit'), rdflib.term.URIRef('https://neuroshapes.org/dash/detailedcircuit/shapes/DetailedCircuitShape'), rdflib.term.URIRef('https://neuroshapes.org/dash/detailedcircuit'), rdflib.term.URIRef('shapes/test_tmp_dir/ontologies/shapes_jsonld_expanded/datashapes/core/detailedcircuit.ttl'))
(rdflib.term.URIRef('bmo:CellTypeAssignmentAnnotation'), rdflib.term.URIRef('https://bbp.epfl.ch/shapes/dash/celltypeassignmentannotation/shapes/CellTypeAssignmentAnnotationShape'), rdflib.term.URIRef('https://bbp.epfl.ch/shapes/dash/celltypeassignmentannotation'), rdflib.term.URIRef('shapes/datashapes/core/celltypeassignmentannotation.json'))
(rdflib.term.URIRef('http://www.w3.org/ns/prov#Collection'), rdflib.term.URIRef('https://neuroshapes.org/dash/collection/shapes/CollectionShape'), rdflib.term.URIRef('https://neuroshapes.org/dash/collection'), rdflib.term.URIRef('shapes/test_tmp_dir/ontologies/bbp/shapes_jsonld_expanded/datashapes/core/collection.ttl'))
(rdflib.term.URIRef('https://neuroshapes.org/Transformation'), rdflib.term.URIRef('https://neuroshapes.org/dash/transformation/shapes/TransformationShape'), rdflib.term.URIRef('https://neuroshapes.org/dash/transformation'), rdflib.term.URIRef('shapes/test_tmp_dir/ontologies/shapes_jsonld_expanded/datashapes/core/transformation.ttl'))
(rdflib.term.URIRef('https://neuroshapes.org/SubjectCollection'), rdflib.term.URIRef('https://neuroshapes.org/dash/subjectcollection/shapes/SubjectCollectionShape'), rdflib.term.URIRef('https://neuroshapes.org/dash/subjectcollection'), rdflib.term.URIRef('shapes/test_tmp_dir/ontologies/bbp/shapes_jsonld_expanded/datashapes/core/subjectcollection.ttl'))
(rdflib.term.URIRef('https://neuroshapes.org/TemplateImageData'), rdflib.term.URIRef('https://neuroshapes.org/dash/templateimagedata/shapes/TemplateImageDataShape'), rdflib.term.URIRef('https://neuroshapes.org/dash/templateimagedata'), rdflib.term.URIRef('shapes/test_tmp_dir/ontologies/bbp/shapes_jsonld_expanded/datashapes/core/templateimagedata.ttl'))
(rdflib.term.URIRef('https://neuroshapes.org/ElectrophysiologyFeature'), rdflib.term.URIRef('https://neuroshapes.org/dash/electrophysiologyfeature/shapes/ElectrophysiologyFeatureShape'), rdflib.term.URIRef('https://neuroshapes.org/dash/electrophysiologyfeature'), rdflib.term.URIRef('shapes/test_tmp_dir/ontologies/bbp/shapes_jsonld_expanded/datashapes/core/electrophysiologyfeature.ttl'))
(rdflib.term.URIRef('https://neuroshapes.org/TraceFeature'), rdflib.term.URIRef('https://neuroshapes.org/dash/tracefeature/shapes/TraceFeatureShape'), rdflib.term.URIRef('https://neuroshapes.org/dash/tracefeature'), rdflib.term.URIRef('shapes/datashapes/core/tracefeature.json'))
(rdflib.term.URIRef('bmo:PSPcapacitanceVoltage'), rdflib.term.URIRef('https://bbp.epfl.ch/shapes/dash/pspcapacitancevoltage/shapes/PSPcapacitanceVoltageShape'), rdflib.term.URIRef('https://bbp.epfl.ch/shapes/dash/pspcapacitancevoltage'), rdflib.term.URIRef('shapes/datashapes/core/pspcapacitancevoltage.json'))
(rdflib.term.URIRef('http://schema.org/QuantitativeValue'), rdflib.term.URIRef('https://neuroshapes.org/dash/quantitativevalue/shapes/QuantitativeValueShape'), rdflib.term.URIRef('https://neuroshapes.org/dash/quantitativevalue'), rdflib.term.URIRef('shapes/test_tmp_dir/ontologies/shapes_jsonld_expanded/datashapes/core/quantitativevalue.ttl'))
(rdflib.term.URIRef('bmo:RawSpikeTrain'), rdflib.term.URIRef('https://bbp.epfl.ch/shapes/dash/rawspiketrain/shapes/RawSpikeTrainShape'), rdflib.term.URIRef('https://bbp.epfl.ch/shapes/dash/rawspiketrain'), rdflib.term.URIRef('shapes/datashapes/core/rawspiketrain.json'))
(rdflib.term.URIRef('bmo:SynapticTypeMapping'), rdflib.term.URIRef('https://bbp.epfl.ch/shapes/dash/synaptictypemapping/shapes/SynapticTypeMappingShape'), rdflib.term.URIRef('https://bbp.epfl.ch/shapes/dash/synaptictypemapping'), rdflib.term.URIRef('shapes/datashapes/core/synaptictypemapping.json'))
(rdflib.term.URIRef('https://neuroshapes.org/NdRaster'), rdflib.term.URIRef('https://neuroshapes.org/dash/ndraster/shapes/NdRasterShape'), rdflib.term.URIRef('https://neuroshapes.org/dash/ndraster'), rdflib.term.URIRef('shapes/datashapes/core/ndraster.json'))
(rdflib.term.URIRef('https://neuroshapes.org/ReconstructedWholeBrainCell'), rdflib.term.URIRef('https://neuroshapes.org/dash/reconstructedwholebraincell/shapes/ReconstructedWholeBrainCellShape'), rdflib.term.URIRef('https://neuroshapes.org/dash/reconstructedwholebraincell'), rdflib.term.URIRef('shapes/test_tmp_dir/ontologies/shapes_jsonld_expanded/datashapes/core/reconstructedwholebraincell.ttl'))
```

crisely09: you can see that many of them are in CURIE form

kgforge/specializations/models/rdf/directory_service.py
```diff
@@ -61,30 +69,47 @@ def resolve_context(self, iri: str) -> Dict:
...
+        class_to_shape = {}
+        shape_to_defining_resource = {}
+        shape_to_named_graph = {}
+        defining_resource_to_named_graph = {}
+        for row in res:
+            class_to_shape[URIRef(row["type"])] = URIRef(row["shape"])
+            shape_to_defining_resource[URIRef(row["shape"])] = URIRef(
+                row["resource_id"]
+            )
+            shape_to_named_graph[URIRef(row["shape"])] = URIRef(row["g"])
```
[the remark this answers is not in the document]
MFSY: So you sort of have:

- Named graph
     - Schema id (== defining resource)
        - Shape id
           - Class id

Removed shape_to_named_graph

kgforge/specializations/models/rdf/directory_service.py
MFSY: Okay.  I made the change. I think this a symptom that the context was not completely taken into account when the directory service is building the graph

kgforge/specializations/models/rdf/service.py
MFSY: > I would expect that even if a resource does not have a type attribute, one could validate if a type_ was provided. 

The current PR is doing that. If type_ is provided then it will be used if not resource.type is used unless it is a list.

kgforge/specializations/models/rdf/service.py
MFSY: Variables are added to config is there is a need to change them and use different values. As of now, the schema data model impose to use OWL.imports, NXV.shapes. So those are not subject to configuration and are fix values. The implementation can evolve though to enable use define schema format but not just yet: when this happen then they ca go to config file

kgforge/specializations/models/rdf/directory_service.py
```diff
@@ -72,15 +72,15 @@ def _build_shapes_map(self) -> Tuple[Dict, Dict, Dict]:
...
         class_to_shape = {}
         shape_to_defining_resource = {}
         defining_resource_to_named_graph = {}
         for row in res:
-            class_to_shape[URIRef(row["type"])] = URIRef(row["shape"])
-            shape_to_defining_resource[URIRef(row["shape"])] = URIRef(
-                row["resource_id"]
-            )
+            shape_uri = URIRef(self.context.expand(row["shape"]))
+            class_to_shape[URIRef(URIRef(self.context.expand(row["type"])))] = shape_uri
```
crisely09, line 82: double URIRef, it should be just once

kgforge/specializations/models/rdf/service.py
crisely09: Yeah, seems like the current RDF model it's already an specialization, it would be nice to make a real `base` class

kgforge/specializations/models/rdf/directory_service.py
ssssarah: I still think the current if elif else structure is not very intuitive. It would be better to have the elif and else in a nested else

kgforge/core/commons/sparql_query_builder.py
ssssarah: I still think the current if elif else structure is not very intuitive. It would be better to have the elif and else in a nested else

kgforge/core/commons/sparql_query_builder.py
ssssarah: It's still a little hard to read. Maybe a multi-line string can help and formatting

```
"""
SELECT (?shape as ?type) ?shape ?resource_id

WHERE { 
    ?shape a sh:NodeShape .
    ?shape a rdfs:Class . 
    %s 
}
""" % ".".join(extra_statements)

File: kgforge/specializations/models/rdf/service.py (hunk 22 above)
Comment (ssssarah) [reply kgforge/specializations/models/rdf/service.py:165]: I wonder why though, I would think in the case of loading the model from Nexus, the named graph and the defining resource would be the same thing

File: kgforge/specializations/models/rdf/store_service.py (hunk 25 above)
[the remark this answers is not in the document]
Comment (ssssarah) [reply kgforge/specializations/models/rdf/store_service.py]: It is the source, and if the source if defined by shape_to_named_graph, then what will tell the implementations apart is how shape_to_named_graph was build = construction, but not which dictionary should be accessed?

File: kgforge/specializations/models/rdf/store_service.py [hunk 28]
```diff
@@ -68,93 +71,95 @@ def resolve_context(self, iri: str) -> Dict:
...
     def generate_context(self) -> Dict:
-        for v in self._shapes_to_resources.values():
-            self._load_shape(v)
+        for shape_uriref, schema_uriref in self.shape_to_defining_resource.items():
+            if str(schema_uriref) not in self._imported:
+                self.transitive_load_shape_graph(
+                    self._get_named_graph_from_shape(shape_uriref), schema_uriref
+                )
         # reloads the shapes graph
         self._sg = ShapesGraphWrapper(self._graph)
```
ssssarah, line 80: This can be changed to self._init_shape_graph_wrapper()

kgforge/specializations/models/rdf/store_service.py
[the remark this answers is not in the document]
ssssarah: We could also consider that the source is the delta endpoint that was targeted to retrieve the schema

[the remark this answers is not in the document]
ssssarah: I think if the delta endpoint option isn't deemed appropriate, there should be a comment in _build_shapes_map of store_service.py explaining the reason for /graph because it's not self-explanatory

kgforge/specializations/models/rdf/service.py
```diff
@@ -272,3 +338,109 @@ def traverse_properties(properties) -> Tuple[Dict, Dict]:
...
+            imported_schema_id = self.context.expand(imported)
+            imported_graph_id = self.defining_resource_to_named_graph[
+                URIRef(imported_schema_id)
+            ]
+            if imported_schema_id not in self._imported:
+                imported_schema_graph = self.transitive_load_shape_graph(
+                    imported_graph_id, imported_schema_id
+                )
+            else:
+                imported_schema_graph = self._graph.graph(URIRef(graph_id))
```
ssssarah, line 357: shouldn't it be self._graph.graph(URIRef(imported_graph_id)) ?

kgforge/specializations/models/rdf/service.py
```diff
@@ -187,23 +220,49 @@ def generate_context(self) -> Dict:
...
+        """Index the loaded node shapes
+        Returns:
+            * a Dict of a targeted owl:Class to a shape
+            * a Dict of a shape to the resource defining it through <https://bluebrain.github.io/nexus/vocabulary/shapes>
+            * a Dict of a resource defining a shape to the a named graph containing it
+        """
+        raise NotImplementedError()
+
+    @abstractmethod
+    def load_shape_graph(self, graph_id: str, schema_id: str) -> Graph:
```
ssssarah, line 233: I think this should have support for graph_id and schema_id being either str of URIRef, and that all code that does self._graph.graph(...) should use this method instead

kgforge/specializations/models/rdf/store_service.py
```diff
@@ -68,93 +71,95 @@ def resolve_context(self, iri: str) -> Dict:
...
+            schema_resource,
+            form="compacted",
+            store_metadata=False,
+            model_context=None,
+            metadata_context=None,
+            context_resolver=self.context_store.service.resolve_context,
+        )
+        # this double conversion was due blank nodes were not "regenerated" with json-ld
+        temp_graph = Graph().parse(data=json.dumps(json_dict), format="json-ld")
+        schema_graph = self._graph.graph(URIRef(graph_id))
```
ssssarah, line 163: I wonder about this one: 
Since we're retrieving the schema, shouldn't we get rid of all the content of schema_graph inside of _graph? 

Especially with the transitive loading, schema_graph probably holds all the imports of this schema too. Might as well override the whole completely, instead of "merging", there could be outdated things?

kgforge/specializations/models/rdf/service.py
```diff
@@ -128,18 +143,34 @@ def lookup_shape_from_node(self, node: URIRef) -> Shape:
 
 class RdfService:
 
-    def __init__(self, graph: Graph, context_iri: Optional[str] = None) -> None:
+    def __init__(self, graph: RDFDataset, context_iri: Optional[str] = None) -> None:
 
         if context_iri is None:
             raise ConfigurationError("RdfModel requires a context")
         self._graph = graph
```
ssssarah, line 150: Wouldn't it be better to have a more evocative name for _graph? Like graph_dataset?

kgforge/specializations/models/rdf/service.py
```diff
@@ -272,3 +338,109 @@ def traverse_properties(properties) -> Tuple[Dict, Dict]:
         context.update({key: terms[key] for key in sorted(terms)})
 
         return {"@context": context} if len(context) > 0 else None
+
+    def transitive_load_shape_graph(
+        self, graph_id: str, schema_id: str, node_shape_uriref: URIRef = None
+    ):
+        schema_graph = self.load_shape_graph(graph_id, schema_id)
+        self._imported.append(schema_id)
```
ssssarah, line 346: I feel like this should only be done when the whole operation is successful, otherwise if an exception is thrown mid-way, the _imported content is not right

kgforge/specializations/models/rdf/service.py
```diff
@@ -272,3 +338,109 @@ def traverse_properties(properties) -> Tuple[Dict, Dict]:
         context.update({key: terms[key] for key in sorted(terms)})
 
         return {"@context": context} if len(context) > 0 else None
+
+    def transitive_load_shape_graph(
```
ssssarah, line 342: Could you document it? For example, what is node_shape_uriref for

kgforge/specializations/models/rdf/service.py
```diff
@@ -272,3 +338,109 @@ def traverse_properties(properties) -> Tuple[Dict, Dict]:
...
+            triples_to_add, _, triples_to_remove = (
+                self._get_property_shapes_from_nodeshape(node_shape_uriref)
+            )
+            for triple_to_add in triples_to_add:
+                schema_graph.add(triple_to_add)
+            for triple_to_remove in triples_to_remove:
+                schema_graph.remove(triple_to_remove)
+        return schema_graph
+
+    def _get_property_shapes_from_nodeshape(self, node_shape_uriref):
```
ssssarah, line 371: Same for this one, documentation?

kgforge/specializations/models/rdf/service.py
```diff
@@ -128,18 +143,34 @@ def lookup_shape_from_node(self, node: URIRef) -> Shape:
...
+        self._sg = None
+        self._init_shape_graph_wrapper()
+        self.NXV = Namespace("https://bluebrain.github.io/nexus/vocabulary/")
         self._context_cache = {}
-        self.classes_to_shapes = self._build_shapes_map()
         resolved_context = self.resolve_context(context_iri)
         self.context = Context(resolved_context, context_iri)
-        self.types_to_shapes: Dict = self._build_types_to_shapes()
+        (
+            self.class_to_shape,
```
ssssarah, line 158: I think for all 3 of those attributes there should be "getters" (for key access) that support both str and URIRef

kgforge/core/commons/sparql_query_builder.py
```diff
@@ -345,20 +367,120 @@ def _replace_in_sparql(
...
+        distinct=True,
+        search_in_graph=search_in_graph,
+        union=True,
+    )
+    if not context:
+        return shacl_query
+    else:
+        return SPARQLQueryBuilder.rewrite_sparql(
+            shacl_query,
+            context_as_dict=context.document,
```
ssssarah, line 483: You should look into the way the read only store uses rewrite_sparql. 

It seems there needs to be a specific shape for context_as_dict to have 
(using _context_to_dict)

kgforge/core/commons/sparql_query_builder.py
```diff
@@ -345,20 +367,120 @@ def _replace_in_sparql(
...
+    shape_target_statement = "?shape sh:targetClass ?type"
+    extra_statements = [defining_statement, deprecated_statement]
+    if statements:
+        extra_statements.extend(statements)
+    shape_target_statement_str = ".".join([shape_target_statement] + extra_statements)
+    shape_node_statements_str = (
+        "SELECT (?shape as ?type) ?shape ?resource_id WHERE { ?shape a sh:NodeShape ."
+        + "?shape a rdfs:Class . {}".format(".".join(extra_statements) + "}")
+    )
+    shape_node_not_targeting_statements_str = (
```
ssssarah, line 458: this statement doesn't seem to return ?type, and then it fails in the tests for 
graph: /tests/data/shacl-model/commons/shapes-3.json
schema: "http://shapes.ex/building"
shape: GeoShape

Not sure if the issue is the shape or the query though

kgforge/specializations/models/rdf/directory_service.py
```diff
@@ -61,30 +64,43 @@ def resolve_context(self, iri: str) -> Dict:
...
+            deprecated_optional=True,
+            search_in_graph=True,
+            context=self.context,
+        )
         res = self._graph.query(query)
-        return {row["type"]: row["shape"] for row in res}
+        class_to_shape = {}
+        shape_to_defining_resource = {}
+        defining_resource_to_named_graph = {}
+        for row in res:
```
ssssarah, line 79: I'm making an assumption here, and I'm not sure, 
But I think inside the SPARQLResult object you iterate over objects of type ResultRow, and because of the shape GeoShape which doesn't have a type, I think none of the result rows have a type (to keep them all uniform, only the intersection of the columns are available), so if "type" in row is always false. 

However if you do
row = row.asdict()
Then "type" in row is True for all shapes but GeoShape.

ssssarah: In fact I wouldn't `if "type" in row`, I would do a regular access surrounded by a try-except block, and maybe raise a custom exception from this.

That is, if "type" is always to be expected from a row (that seems to be the reason by the tests are failing). If "type" is not always expected, then discard this comment

tests/specializations/models/test_rdf_directory_service.py
```diff
@@ -0,0 +1,60 @@
...
+        expected_g = rdflib.Graph().parse(file_path, format="json-ld")
+        assert len(g) == len(expected_g)
+
+
+def test_build_shapes_map(rdf_model_from_dir: RdfModel):
+    (class_to_shape, shape_to_defining_resource, defining_resource_to_named_graph) = (
+        rdf_model_from_dir.service._build_shapes_map()
+    )
+
+    assert sorted(list(class_to_shape.values())) == sorted(
```
ssssarah, line 46: I guess if we accept that some shapes will target no class (which is the case for GeoShape), this assertion will not be True

kgforge/specializations/models/rdf/service.py
```diff
@@ -272,3 +343,121 @@ def traverse_properties(properties) -> Tuple[Dict, Dict]:
...
+                property_shapes_to_add.append((SH.property, sh_node_property))
+
+        return triples_to_add, property_shapes_to_add, triples_to_remove
+
+    def get_shape_graph(self, shape_uriref: URIRef) -> Tuple[Shape, Graph]:
+        try:
+            shape = self.get_shape_graph_wrapper().lookup_shape_from_node(shape_uriref)
+            if (
+                shape_uriref in self.shape_to_defining_resource
+                and self.shape_to_defining_resource[shape_uriref] in self._imported
```
ssssarah, line 433: I wonder about the reason for the existence of the variable self._imported.
I would think that checking the states of the dictionaries or the graph dataset could let us know about whether something is imported, but the maintenance of this extra array may lead to things not being in sync (is it a valid state to have an entry in self.shape_to_defining_resource, but not to be in self._imported, or the other way around?)

ssssarah: I see that self._imported.append() only happens after `self.load_shape_graph` and inside it creates one graph into `_graph`. So maybe there can be a method called is_imported that checks whether there is an entry inside `self._graph` for a named graph using for `defining_resource_to_named_graph[schema_uriref]`

ssssarah: Then it depends what the definition of imported is, does that also include transitive imports?

kgforge/specializations/models/rdf/store_service.py
MFSY: Very unlikely that something is in the schema_graph at this point before parsing. But I added schema_graph.remove((None, None, None)) to empty it before parsing.

kgforge/core/commons/sparql_query_builder.py
MFSY: for the query built by build_shacl_query, context.document is enough as only adding sh and rdfs prefixes are needed. Now the jsonld context processing logic in [here](https://github.com/BlueBrain/nexus-forge/blob/master/kgforge/core/archetypes/read_only_store.py#L58) should be brought back in SPARQLQueryBuilder.rewrite_sparql which should accept a Context

kgforge/core/commons/sparql_query_builder.py
MFSY: This should be fix now

kgforge/specializations/models/rdf/directory_service.py
MFSY: I go with `if row["type"]` which work/

tests/specializations/models/test_rdf_directory_service.py
MFSY: Yes. The test is adapted now

kgforge/specializations/models/rdf/service.py
MFSY: I moved it to the end of the method

kgforge/specializations/models/rdf/service.py
MFSY: I moved it to the end of this method

kgforge/specializations/models/rdf/service.py
MFSY: 4 calls actually:

1. 1 in service._transitive_load_shape_graph
2. 1 in service.get_shape_graph
3. 1 in directory_service.load_shape_graph
4. 1 in store_service.load_shape_graph

service.get_shape_graph is the method to call to get a shape graph from a shape URI (not load_shape_graph).
load_shape_graph is for loading a shape in service._graph from  source (dir or store). To clarify, I renamed it to load_shape_graph_from_source. 

service.get_shape_graph  can't be used in 1, 3 and 4 because service.get_shape_graph actually uses them. These services are internal to the RDFModel class, when these features to get shape graph will be added to RDFModel then we can support Union of type

kgforge/specializations/models/rdf/service.py
MFSY: Doc added

kgforge/specializations/models/rdf/service.py
MFSY: Doc added

kgforge/specializations/models/rdf/service.py
MFSY: Let go for _dataset_graph

kgforge/specializations/models/rdf/service.py
MFSY: Reevaluate this. I think at service level adding getters would bring overhead.

kgforge/specializations/models/rdf/service.py
MFSY: Once we have another non demo model we can evaluate the need for a base non core model class

kgforge/core/commons/sparql_query_builder.py
MFSY: Not sure to understand this. statements can be list or not when not union: no need for raising exception.

kgforge/specializations/models/rdf/store_service.py
ssssarah: There's an unused import resulting from that (ShapesGraphWrapper doesn't need to be imported)

kgforge/specializations/models/rdf/service.py
ssssarah: I meant that there are multiple calls to self._dataset_graph.graph(...) that can be replaced with load_shape_graph_from_source, they all start from a graph id. 
In some cases it's a str, in others it's already a URIRef, so it would be best to call load_shape_graph_from_source instead of those, and have that function support both str and uriref as input

Screenshot 2024-03-25 at 14 51 13

kgforge/core/commons/sparql_query_builder.py
ssssarah: if not union:
else: 
  if  isinstance(statements, list):
  else: 

is more intuitive

kgforge/specializations/models/rdf/service.py
```diff
@@ -272,3 +350,191 @@ def traverse_properties(properties) -> Tuple[Dict, Dict]:
...
+            if (
+                node_shape_uriref in self.shape_to_defining_resource
+                and self.shape_to_defining_resource[node_shape_uriref] in self._imported
+            ):
+                shape_graph = self._dataset_graph.graph(
+                    self._get_named_graph_from_shape(node_shape_uriref)
+                )
+            else:
+                raise ValueError()
+        except Exception:
```
ssssarah, line 500: This is quite broad, what can cause this? 

Can you explain what goes on in terms of logic, what sequence of events will go through the try, and what sequence of events will go through the except?

kgforge/specializations/models/rdf/service.py
```diff
@@ -241,7 +237,7 @@ def _build_shapes_map(self) -> Tuple[Dict, Dict, Dict]:
 
     @abstractmethod
     def load_shape_graph_from_source(self, graph_id: str, schema_id: str) -> Graph:
-        """Loads into graph_id the node shapes defined in shema_id
+        """Loads into graph_id the node shapes defined in shema_id from the source
```
ssssarah, line 240: little typo

PR #397, opened 2024-04-23
fix search when looking for a literal to be unequal to a value

ssssarah:
When specifying that a value which is a literal that should be unequal to a value, it would check that it is equal to it. This fixes that


kgforge/core/commons/sparql_query_builder.py
```diff
@@ -169,20 +169,15 @@ def build(
...
-                            )
-                        statements.append(f"{property_path} ?v{index}")
-                        sparql_filters.append(
-                            f"FILTER(?v{index} = {_box_value_as_full_iri(value)})"
-                        )
-                    else:
-                        statements.append(f"{property_path} ?v{index}")
-                        sparql_filters.append(
-                            f"FILTER(?v{index} {sparql_operator_map[f.operator]} {_box_value_as_full_iri(value)})"
+                    if value_type is CategoryDataType.LITERAL and f.operator not in ["__eq__", "__ne__"]:
```
crisely09, line 172: do we have a test for this? checking if it is a string, it should only accept == or !=

MFSY: There is a test in https://github.com/BlueBrain/nexus-forge/blob/master/tests/specializations/stores/test_bluebrain_nexus.py#L560

crisely09 (dismissed):
looks good

kgforge/core/commons/sparql_query_builder.py
MFSY: I am wandering though why this test were passing. Looking at the change it might mean we did not have a Literal value type in the test. @ssssarah can add that case in the test ?

PR #406, opened 2024-05-24
Add method when initializing forge to export pyshacl environment variable

crisely09:
To ensure the pyshacl environment variable is set when using `forge.validate`, I added a method to set it at initializing a `KnowledgeGraphForge` instance.


AurelienJaquier (approved):
Great! Thank you @crisely09 !

MFSY (approved):

PR #407, opened 2024-05-29
rm nexus sdk usage and dependency

ssssarah:
I've put all the code from the sdk that was used in the file previous_sdk_usage, with the change that the endpoint and token are provided as parameters (as opposed to setting environment variables)

No design thought really went into this, it's a fast solution, up to change

(preparatory work so that we specify a token as a callable, without having to handle setting the nexussdk environment variable)


kgforge/specializations/stores/bluebrain_nexus.py
lecriste, line 1: Shall we have a test for the function `_upload_one` changed here?

ssssarah: Sure, but we don't have integration tests in this repository so I wouldn't know how to test it?

lecriste (approved):

kgforge/specializations/stores/bluebrain_nexus.py
crisely09: This is sort of "tested" with the notebooks, the files are used in the getting started notebooks. Just check if they run correctly.

kgforge/specializations/stores/nexus/previous_sdk_usage.py
```diff
@@ -0,0 +1,297 @@
+import puremagic
```
crisely09, line 1: what are the differences with mime types?

crisely09:
hmm why to call things: `previous_sdk_usage.py` ? why not just something generic about the methods there? and just comment/docstring that this was previois sdk usage

kgforge/specializations/stores/bluebrain_nexus.py
crisely09: I checked, notebooks run fine.

PR #413, opened 2024-06-21
rewrite uri as static method

ssssarah:
For the sake of being able to rewrite the URI without having a forge instance tied to the appropriate bucket


crisely09 (approved):
looks good to me

Issue #414, opened 2024-06-26
Use schema endpoint to register schemas


crisely09:
Schemas, as well as views, resolvers, etc. are especial types of resources, they have a separate endpoint in delta API.
The idea is to detect the schema_id: `https://bluebrain.github.io/nexus/schemas/shacl-20170720.ttl` and use a private method `_register/update/deprecate_schema` pointing to the schema endpoint.

crisely09:
Solved with #415

PR #415, opened 2024-06-28
Check for schema id to use schema endpoint

crisely09:
To address #414

Add a method to build the url checking if the schema_id is the SHACL schema.


kgforge/specializations/stores/nexus/service.py
```diff
@@ -210,6 +212,12 @@ def make_endpoint(endpoint: str, endpoint_type: str, organisation: str, project:
             (endpoint, endpoint_type, quote_plus(organisation), quote_plus(project))
         )
 
+    @staticmethod
+    def add_resource_id_to_endpoint(
+            endpoint: str, resource_id: Optional[str]
+    ):
+        return "/".join([endpoint, quote_plus(resource_id)])
```
lecriste, line 219: Since `resource_id` is Optional, don't you need to check that it's different from `None` before using it?

crisely09: oh yes, missed that, fixed in the last commit

lecriste (approved):

PR #418, opened 2024-07-10
Parallelize validate

ssssarah:
- the traverse function being optionally overwritten on Shape was due to a change in pyshacl versions (I believe an old version had this version but a newer one no longer had it. For the version we have fixed, this function does not exist) https://github.com/BlueBrain/nexus-forge/issues/43 https://github.com/BlueBrain/nexus-forge/pull/44 
- renamed usage of `_transitive_load_shape_graph` to `_transitive_load_resource_graph`
- dummy implementation of `_validate_many` in `DemoModel`
- Override pyshacl `ShapesGraph` implementation of `def shapes` to return `list` instead of `dict_values()` so that it can be picklable (for multiprocessing)
- Parallelized validate. 
  -> cannot use threads because pyshacl uses rdflib, rdflib uses pyparsing and it is not thread safe
  -> shape loading occurs before creating multiple processes for validation. It would be possible to load shapes in individual processes, however: 
       - "caching" of shapes becomes useless, the same shape will be queried multiple times across processes 
       - Shape loading in individual processes would be best if validating resources of different types/shapes
       - The former point might not be completely true because shapes re-use each other
       - It is for sure better to load the shapes beforehand if we're dealing with a list of resources of the same type/shape.
    Given these reasons, I opted for loading all shapes beforehand.

### DIFF
--- a/kgforge/specializations/models/rdf/service.py
+++ b/kgforge/specializations/models/rdf/service.py
@@ -63,52 +63,51 @@ ALL_COLLECTORS = [
 ALL_COLLECTORS_MAP = {c.constraint(): c for c in ALL_COLLECTORS}
 
 
-def traverse(self, predecessors: Set[URIRef]) -> Tuple[List, Dict]:
-    """traverses the Shape SACL properties to collect constrained properties
-
-    This function is injected to pyshacl Shape object in order to traverse the Shacl graph.
-    It will call a specific collector depending on the SHACL property present in the NodeShape
-
-    Args:
-        predecessors: list of nodes that have being traversed, used to break circular
-            recursion
-
-    Returns:
-        properties, attributes: Tuple(list,dict), the collected properties and attributes
-            respectively gathered from the collectors
-    """
-
-    parameters = self.parameters()
-    properties = []
-    attributes = {}
-    done_collectors = set()
-    for param in iter(parameters):
-        if param in ALL_COLLECTORS_MAP:
-            constraint_collector = ALL_COLLECTORS_MAP[param]
-            if constraint_collector not in done_collectors:
-                c = constraint_collector(self)
-                predecessors.add(self.node)
-                props, attrs = c.collect(predecessors)
-                if attrs:
-                    attributes.update(attrs)
-                if props:
-                    properties.extend(props)
-                done_collectors.add(constraint_collector)
-                if predecessors:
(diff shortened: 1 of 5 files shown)


ssssarah:
Unfortunately I ran into an issue when testing validation on 4000 resources. 
Long story short, pyshacl uses rdflib, rdflib uses pyparsing and it is not thread safe 
https://github.com/RDFLib/rdflib/issues/765 I'll look more into it

kgforge/specializations/models/rdf_model.py
```diff
@@ -127,33 +130,72 @@ def validate(
...
+
+        resource._validated = conforms
+        resource._last_action = Action(operation="_validate_many", succeeded=conforms, error=err)
+
+        if not conforms and raise_:
+            raise err
+
+        return resource
+
+    def _get_shape_stuff(self, r: Resource, type_: str) -> Tuple[str, ShapeWrapper, Graph, Graph, Resource]:
```
crisely09, line 163: not convinced about this name

kgforge/specializations/models/rdf/service.py
```diff
@@ -138,12 +137,19 @@ def lookup_shape_from_node(self, node: URIRef) -> Shape:
...
-            shape_wrapper = ShapeWrapper(self._node_shape_cache[node])
-            if not hasattr(shape_wrapper, "traverse"):
-                shape_wrapper.traverse = types.MethodType(traverse, shape_wrapper)
-            return shape_wrapper
-        return shape
+
+        return ShapeWrapper(shape)
+
+    @property
+    def shapes(self):  # pyshacl implementation returns dict_values (not list). This cannot be pickled.
```
crisely09, line 144: maybe `shapes_list` would be even more accurate?

kgforge/specializations/models/rdf_model.py
ssssarah: that is fair 😅

kgforge/specializations/models/rdf/service.py
ssssarah: This is overriding a property defined by pyshacl, so I have to stick to the same name

crisely09: Ah ok

crisely09 (approved):

PR #425, opened 2024-10-31
Use jsonpath to collect values when downloading a file

crisely09:
There was a problem when doing something like:
```
forge.download(cell_types_ontology, path=my_path, content_type='application/ld+json')
```
when the encodingFormat was JSON,  as it was used in the headers, and the metadata of the file was return instead of the content of the file.

Solution:
Use the default headers (`Accept : */*`) to download the content of a file and switch to jsonpath to collect values from a given path.


lecriste (approved):
LGTM

